PR #1, opened 2023-10-19
Register custom drivers properly

yhabteab:


lippserd (changes requested):
I'm fine with the renames but not with the automatic driver registration. See here pls:

https://github.com/Icinga/icingadb/blame/library-split/cmd/icingadb/main.go#L62

lippserd (approved):

lippserd:
Main branch was force pushed because the latest version of the Icinga Go library and the library code in the Icinga DB have diverged. This change is obsolete now.

PR #2, opened 2023-10-19
Add enhancements used by `notifications` daemon

yhabteab:


lippserd:

PR #5, opened 2023-12-11
database: Introduce `ColumnMap`

lippserd:
`ColumnMap` provides a cached mapping of structs exported fields to their database column names. By default, all exported struct fields are mapped to their database column names using snake case notation. The `-` (hyphen) directive for the db tag can be used to exclude certain fields. Since `ColumnMap` uses cache, the returned slice MUST NOT be modified directly.

### DIFF
--- a/database/column_map.go
+++ b/database/column_map.go
@@ -1,0 +1,75 @@
+package database
+
+import (
+	"database/sql/driver"
+	"github.com/jmoiron/sqlx/reflectx"
+	"reflect"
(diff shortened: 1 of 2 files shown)


lippserd:
Main branch was force pushed because the latest version of the Icinga Go library and the library code in the Icinga DB have diverged. This change is now reflected by commit 102481adcdfcc54583eaf8627dad4d2be818df18.

PR #7, opened 2024-01-19
Allow to dynamically define type constraint name

yhabteab:
This allows us to dynamically determine the constraint names used for the `UPSERT` and `INSERT IGNORE` statements for the PostgreSQL driver.

### DIFF
--- a/database/contracts.go
+++ b/database/contracts.go
@@ -47,3 +47,10 @@ type TableNamer interface {
 type Scoper interface {
 	Scope() any
 }
+
+// PgsqlOnConflictConstrainter implements the PgsqlOnConflictConstraint method,
+// which returns the primary or unique key constraint name of the PostgreSQL table.
+type PgsqlOnConflictConstrainter interface {
+	// PgsqlOnConflictConstraint returns the primary or unique key constraint name of the PostgreSQL table.
+	PgsqlOnConflictConstraint() string
+}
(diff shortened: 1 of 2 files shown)


lippserd (changes requested):
Since this is PostgreSQL specific I would reflect that in the interface/function name. I know it's rather long but what about something like `PostgresOnConflictConstraint()`?

database/contracts.go
```diff
@@ -47,3 +47,9 @@ type TableNamer interface {
 type Scoper interface {
 	Scope() any
 }
+
+// PgsqlOnConflictConstrainter implements the PgsqlOnConflictConstraint method,
+// which returns the PK or unique key constraint name of the PostgreSQL table.
```
lippserd, line 52:
```suggestion
// which returns the primary or unique key constraint name of the PostgreSQL table.
```

lippserd (changes requested):

database/contracts.go
```diff
@@ -47,3 +47,9 @@ type TableNamer interface {
...
+type PgsqlOnConflictConstrainter interface {
+	PgsqlOnConflictConstraint() string // PgsqlOnConflictConstraint returns the PK/UNIQUE key constraint name
```
lippserd, line 54:
```suggestion
        // PgsqlOnConflictConstraint returns the primary or unique key constraint name of the PostgreSQL table.
        PgsqlOnConflictConstraint() string
```

lippserd (changes requested):

lippserd (approved):
Please feel free to merge.

lippserd:
Main branch was force pushed because the latest version of the Icinga Go library and the library code in the Icinga DB have diverged. This change is now reflected by commit 54491de50190ee9517a2228f4a43053f679145c9.

PR #8, opened 2024-01-22
Add go github workflows

yhabteab:


.github/workflows/go.yml
```diff
@@ -0,0 +1,29 @@
...
+        run: go vet ./...
+      - name: Go linter
+        uses: golangci/golangci-lint-action@v3
+        with:
+          version: latest
+          only-new-issues: true
+
+          # Enable the gosec linter w/o having to create a .golangci.yml config
+          args: -E gosec
+          skip-cache: true
```
lippserd, line 29:
```suggestion
jobs:

  go:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - name: Set up Go
        uses: actions/setup-go@v5
        with:
          go-version: stable
          cache: false

      - name: Lint
        uses: golangci/golangci-lint-action@v3
        with:
          version: latest
          only-new-issues: true

          # Enable the gosec linter w/o having to create a .golangci.yml config
          args: -E gosec

      - name: Test
        run: go test -v ./...
```

I quite don't like the newline situation and there's no need to run `go vet` as it's [enabled by default](https://golangci-lint.run/usage/linters/#enabled-by-default) in `golangci-lint`. Also, `golangci-lint` claims to have [superior caching](https://github.com/golangci/golangci-lint-action?tab=readme-ov-file#performance), so let's try that and run tests last.

lippserd (changes requested):

lippserd:
Main branch was force pushed because the latest version of the Icinga Go library and the library code in the Icinga DB have diverged. This change is now reflected by commit 2b013450a107e28ca5ca5dfe8035cd321154f73e.

PR #9, opened 2024-02-23
workflows: Fix `gosec` errors & enable code annotations

yhabteab:
Enables code annotations in case of failures and fixes the actions triggered by the main branch.
https://github.com/Icinga/icinga-go-library/actions/runs/7783022809/job/21220618068

**Edit:**
This PR suppresses some of the reported existing `gosec` errors and solves some of them as well and drops the `only-new-issue` flag set for the go linter.


yhabteab:
If these errors do not get fixed with the new permissions, we may have to either fix or suppress them.
https://github.com/Icinga/icinga-go-library/actions/runs/7783022809/job/21220618068

config/config.go
```diff
@@ -80,7 +81,7 @@ func (t *TLS) MakeConfig(serverName string) (*tls.Config, error) {
 		return nil, nil
 	}
 
-	tlsConfig := &tls.Config{}
+	tlsConfig := &tls.Config{} // #nosec G402 -- TLS MinVersion too low - we can't abandon TLS 1.0 as long as Centos 7 is alive.
```
julianbrost, line 84: CentOS 7 isn't a reason for that. The OpenSSL version in the base system only doesn't support TLS 1.3, TLS 1.2 works just fine there.

yhabteab: Does that mean you want to set the min required TLS version to 1.2?

oxzi: But isn't this code being used to create a TLS configuration to access a database or a Redis server? Thus, does, i.e., a MariaDB on CentOS 7 support TLS 1.2?

julianbrost: > Does that mean you want to set the min required TLS version to 1.2?

Primarily, I wanted to doubt the claim that "we can't abandon TLS 1.0 as long as Centos 7 is alive".

In general, TLS 1.2 is a reasonable minimal requirement, everything older is better avoided unless there's a very good reason why you would need it (and SSL is a big no right away). And CentOS/RHEL 7, don't seem to be such a reason, the [RHEL 7 documentation already recommends TLS 1.2](https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/security_guide/sec-hardening_tls_configuration#tab-TLS-Protocol_Versions).

lippserd:

PR #10, opened 2024-04-05
Support to post initialise sql connections

yhabteab:
refs https://github.com/Icinga/icingadb/pull/711


database/set_mysql_session_vars_test.go
```diff
@@ -0,0 +1,63 @@
...
+		},
+		"VariablesWithInvalidValues": { // System variables set to an invalid value
+			{name: "transaction_isolation", value: "REPEATABLE-WRITE", expect: &mysql.MySQLError{Number: 1231}},
+			{name: "completion_type", value: int64(-10), expect: &mysql.MySQLError{Number: 1231}}, // Min valid val 0
+			{name: "completion_type", value: int64(10), expect: &mysql.MySQLError{Number: 1231}},  // Max valid val 2
+			{name: "explain_format", value: "IcingaDB", expect: &mysql.MySQLError{Number: 1231}},
+		},
+	}
+
+	rdb := it.MysqlDatabase()
```
julianbrost, line 42: We don't want to make the `icinga-go-library` module depend on `icinga-testing` as that would mean that all its dependencies would be pulled into its users, including Docker. That's one of the reasons why `tests/` in Icinga DB is a separate Go module so that these dependencies are kept separate from the main module.

Note that I also would take the opportunity to add tests for more of the database code. Like starting a database externally, passing the connection details to the tests, for example in environment variables, then filling a database config struct with these values, and use this to connect. That would then serve as the basis for more tests in this project and also avoid the aforementioned dependency.

julianbrost (changes requested):

lippserd:

PR #12, opened 2024-05-23
Introduce database.AutoUpgradeSchema()

Al2Klimov:


Al2Klimov:
@lippserd Does the design look good so far?

lippserd:

PR #17, opened 2024-05-27
Introduce Query builder

### LINKED ISSUE #57: Adjustable Queries
# Requirements

* Not all columns may be used in queries.
* `WHERE` clause used in statements must be adjustable. Default for `DELETE` for example is `id IN (?)` at the moment.
* Executed query should be replaceable, e.g. `ON DUPLICATE KEY` may specify `col = col + VALUES(col)`.
* All functionality shall be usable w/o building custom types, e.g. table name, columns as `string`, `[]string`.

## Possible API

```go
// Using query builder directly
sql := NewQueryBuilder(db, subject).SetColumns(columns).Insert()
db.NamedBulkExec(ctx, sql, ch)

// Functional options
db.Upsert(ctx, ch, options.SetColumns(columns))

// Using query builder directly #2
upsert := qb.Upsert(subject).Columns(columns)
db.NamedBulkExec(ctx, upsert, ch)
// or
upsert.Stream(ctx, db, ch)

db.Upsert().Columns(columns).Exec() // .Stream()

qb.Upsert().Columns().Exec(ctx, db, setofData) // .Stream(ctx, db, ch)
```

yhabteab:
resolves #57
refs #16 (2a133da is cherry-picked from that PR)

### DIFF
--- a/database/query_builder.go
+++ b/database/query_builder.go
@@ -1,0 +1,244 @@
+package database
+
+import (
+	"fmt"
+	"golang.org/x/exp/slices"
+	"reflect"
+	"sort"
+	"strings"
+)
+
+// QueryBuilder is an addon for the [DB] type that takes care of all the database statement building shenanigans.
+// The recommended use of QueryBuilder is to only use it to generate a single query at a time and not two different
+// ones. If for instance you want to generate `INSERT` and `SELECT` queries, it is best to use two different
+// QueryBuilder instances. You can use the DB#QueryBuilder() method to get fully initialised instances each time.
+type QueryBuilder struct {
+	subject         any
+	columns         []string
+	excludedColumns []string
+
+	// Indicates whether the generated columns should be sorted in ascending order before generating the
+	// actual statements. This is intended for unit tests only and shouldn't be necessary for production code.
+	sort bool
+}
+
+// SetColumns sets the DB columns to be used when building the statements.
+// When you do not want the columns to be extracted dynamically, you can use this method to specify them manually.
+// Returns the current *[QueryBuilder] receiver and allows you to chain some method calls.
+func (qb *QueryBuilder) SetColumns(columns ...string) *QueryBuilder {
+	qb.columns = columns
+	return qb
+}
+
+// SetExcludedColumns excludes the given columns from all the database statements.
+// Returns the current *[QueryBuilder] receiver and allows you to chain some method calls.
+func (qb *QueryBuilder) SetExcludedColumns(columns ...string) *QueryBuilder {
+	qb.excludedColumns = columns
+	return qb
+}
+
+// Delete returns a DELETE statement for the query builders subject filtered by ID.
+func (qb *QueryBuilder) Delete() string {
+	return qb.DeleteBy("id")
+}
+
+// DeleteBy returns a DELETE statement for the query builders subject filtered by the given column.
+func (qb *QueryBuilder) DeleteBy(column string) string {
+	return fmt.Sprintf(`DELETE FROM "%s" WHERE "%s" IN (?)`, TableName(qb.subject), column)
+}
+
+// Insert returns an INSERT INTO statement for the query builders subject.
+func (qb *QueryBuilder) Insert(db *DB) (string, int) {
+	columns := qb.BuildColumns(db)
+
+	return fmt.Sprintf(
+		`INSERT INTO "%s" ("%s") VALUES (:%s)`,
+		TableName(qb.subject),
+		strings.Join(columns, `", "`),
+		strings.Join(columns, ", :"),
+	), len(columns)
+}
+
+// InsertIgnore returns an INSERT statement for the query builders subject for
+// which the database ignores rows that have already been inserted.
+func (qb *QueryBuilder) InsertIgnore(db *DB) (string, int) {
+	columns := qb.BuildColumns(db)
+
+	var clause string
+	switch db.DriverName() {
+	case MySQL:
+		// MySQL treats UPDATE id = id as a no-op.
+		clause = fmt.Sprintf(`ON DUPLICATE KEY UPDATE "%[1]s" = "%[1]s"`, columns[0])
+	case PostgreSQL:
+		clause = fmt.Sprintf("ON CONFLICT ON CONSTRAINT %s DO NOTHING", qb.getPgsqlOnConflictConstraint())
+	default:
+		panic("Driver unsupported: " + db.DriverName())
+	}
+
+	return fmt.Sprintf(
+		`INSERT INTO "%s" ("%s") VALUES (:%s) %s`,
+		TableName(qb.subject),
+		strings.Join(columns, `", "`),
+		strings.Join(columns, ", :"),
+		clause,
+	), len(columns)
+}
+
+// Select returns a SELECT statement from the query builders subject and the already set columns.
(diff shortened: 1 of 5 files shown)


database/query_builder.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// Indicates whether the generated columns should be sorted in ascending order before generating the
+	// actual statements. This is intended for unit tests only and shouldn't be necessary for production code.
+	sort bool
+}
+
+// NewQB returns a fully initialized *QueryBuilder instance for the given subject/struct.
+func NewQB(subject any) *QueryBuilder {
+	return &QueryBuilder{subject: subject}
+}
```
julianbrost, line 28: I wouldn't abbreviate it here. `NewQueryBuilder` still has a reasonable length for a function name.

Also, doesn't pretty much every use of it require access to `*DB` in the end? So another option could be `func (db *DB) QueryBuilder(subject any) *QueryBuilder`.

oxzi: > I wouldn't abbreviate it here. `NewQueryBuilder` still has a reasonable length for a function name.

I would second this.

> Also, doesn't pretty much every use of it require access to `*DB` in the end? So another option could be `func (db *DB) QueryBuilder(subject any) *QueryBuilder`.

Even as this makes sense for most practical use cases, I would prefer to separate a query builder (which only builds queries, by its name) from a database connection. This also eases testing and potential mocking.

julianbrost: > Even as this makes sense for most practical use cases, I would prefer to separate a query builder (which only builds queries, by its name) from a database connection. This also eases testing and potential mocking. Or, in other words, keep things as they currently are.

The struct type `QueryBuilder` remains, so that can still be used in tests directly. It could also have a member of a type like `interface { DriverName() string }` (that should be everything it needs) and the method on `*DB` would just be a convenience function for normal use. Tests could then initialize it with a dummy type that implements the same interface.

database/query_builder_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
+	"testing"
+	"time"
+)
+
+func TestQueryBuilder(t *testing.T) {
+	t.Parallel()
+
+	c := &Config{}
```
oxzi, line 16: I'd also like to see tests for PostgreSQL as there are some differences which should at least be tested.

database/query_builder.go
```diff
@@ -0,0 +1,260 @@
...
+	var columns []string
+	if t.Kind() == reflect.String {
+		columns = []string{subject.(string)}
+	} else if t.Kind() == reflect.Struct || t.Kind() == reflect.Pointer {
+		if scoper, ok := subject.(Scoper); ok {
+			return qb.Where(db, scoper.Scope())
+		}
+
+		columns = db.columnMap.Columns(subject)
+	}
```
oxzi, line 218: Should the `else` case also be considered by a `panic` or silently be dropped as currently done because this shouldn't happen™.

database/query_builder.go
```diff
@@ -0,0 +1,254 @@
...
+			return false
+		})
+	}
+
+	if qb.sort {
+		// The order in which the columns appear is not guaranteed as we extract the columns dynamically
+		// from the struct. So, we've to sort them here to be able to test the generated statements.
+		sort.SliceStable(columns, func(a, b int) bool {
+			return columns[a] < columns[b]
+		})
```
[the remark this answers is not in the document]
julianbrost: With that being the default comparison of `string`, just `slices.Sort(columns)` should do (alternatively, there's even the pre-generics [`sort.Strings()`](https://pkg.go.dev/sort#Strings))

database/query_builder.go
```diff
@@ -0,0 +1,254 @@
...
+		strings.Join(set, ", "),
+	), len(insertColumns)
+}
+
+// Where returns a WHERE clause with named placeholder conditions built from the
+// specified scoper/column combined with the AND operator.
+func (qb *QueryBuilder) Where(subject any) (string, int) {
+	t := reflect.TypeOf(subject)
+	if t == nil { // Subject is a nil interface value.
+		return "", 0
```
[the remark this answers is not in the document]
yhabteab: Sorry, but if you continue processing even though `‘""` and `0` are returned, then you are responsible for that, but not this method. Panicking here would cause such existing code to break, and actually `nil` is not that disastrous to me and should not cause the method to panic.
https://github.com/Icinga/icinga-go-library/blob/474ff254aab448c828bd1c15dfa1988e05a6fce8/database/query_builder.go#L93-L96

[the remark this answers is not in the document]
oxzi: Okay, than never mind :>

database/query_builder.go
```diff
@@ -0,0 +1,254 @@
...
+func (qb *QueryBuilder) BuildColumns() []string {
+	var columns []string
+	if len(qb.columns) > 0 {
+		columns = qb.columns
+	} else {
+		columns = qb.db.columnMap.Columns(qb.subject)
+	}
+
+	if len(qb.excludedColumns) > 0 {
+		columns = slices.DeleteFunc(append([]string(nil), columns...), func(column string) bool {
```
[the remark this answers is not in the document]
yhabteab: > By doing so, the slices.DeleteFunc does not alter the referenced columns = qb.columns original slice. 

Actually, I wouldn't pay that much attention if this was the only rationale for this, but since `qb.columns` may contain columns mapped from the internal DB mapper, which explicitly warns you not to change the returned slice directly, I do it unconditionally.
https://github.com/Icinga/icinga-go-library/blob/2d47d95fe9d8d3d43d359de8569d9abdf6fecb9d/database/column_map.go#L12-L16

lippserd:
I would find it clearer if the API did not require the query builder to be created again and again. It would also be more flexible if the different statement types were separate structs.

julianbrost:
If we're starting to discuss the overall API design this PR won't do it to get https://github.com/Icinga/icinga-notifications/pull/114 merged any time soon. So either we merge this as an experimental API (i.e. might fundamentally change) or we need another PR that simply exposes a `BuildColumns()` function again so that this can be used by https://github.com/Icinga/icinga-notifications/pull/114 in the meantime.

lippserd:
> If we're starting to discuss the overall API design this PR won't do it to get [Icinga/icinga-notifications#114](https://github.com/Icinga/icinga-notifications/pull/114) merged any time soon. So either we merge this as an experimental API (i.e. might fundamentally change) or we need another PR that simply exposes a `BuildColumns()` function again so that this can be used by [Icinga/icinga-notifications#114](https://github.com/Icinga/icinga-notifications/pull/114) in the meantime.

I see. Then let’s just merge it.

oxzi (dismissed):

julianbrost:
> or we need another PR that simply exposes a `BuildColumns()` function again so that this can be used by [Icinga/icinga-notifications#114](https://github.com/Icinga/icinga-notifications/pull/114) in the meantime

I've asked for this to be done nonetheless as I find the current interface this PR adds more confusing than it needs to be in regards to what it accepts in `any` parameters. And even if it's kept that way, the current implementation has room for improvement by reducing the amount of use of the `reflect` package. All in all, more than what should be necessary to finally get Icinga Notifications away from importing Icinga DB as a package.

PR #18, opened 2024-05-27
Add query execution utilities

yhabteab:


database/utils.go
```diff
@@ -43,6 +45,49 @@ func SplitOnDupId[T IDer]() com.BulkChunkSplitPolicy[T] {
...
+		query := stmt + " RETURNING id"
+		ps, err := conn.PrepareNamedContext(ctx, query)
+		if err != nil {
+			return types.Int{}, errors.Wrapf(err, "cannot prepare %q", query)
+		}
+		// We're deferring the ps#Close invocation here just to be on the safe side, otherwise it's
+		// closed manually later on and the error is handled gracefully (if any).
+		defer func() { _ = ps.Close() }()
+
+		if err = ps.GetContext(ctx, &resultID, arg); err != nil {
```
Al2Klimov, line 67: err can be a new variable w/o changing its name. TL;DR: Please use :=.

database/utils.go
```diff
@@ -43,6 +45,49 @@ func SplitOnDupId[T IDer]() com.BulkChunkSplitPolicy[T] {
...
+	case MySQL:
+		result, err := sqlx.NamedExecContext(ctx, conn, stmt, arg)
+		if err != nil {
+			return types.Int{}, CantPerformQuery(err, stmt)
+		}
+
+		resultID, err = result.LastInsertId()
+		if err != nil {
+			return types.Int{}, errors.Wrap(err, "cannot retrieve last inserted ID")
+		}
```
Al2Klimov, line 83: I mean, you could also make this the default branch.

database/contracts.go
```diff
@@ -54,3 +59,11 @@ type PgsqlOnConflictConstrainter interface {
...
 	PgsqlOnConflictConstraint() string
 }
+
+// TxOrDB is just a helper interface that can represent a *[sqlx.Tx] or *[DB] instance.
+type TxOrDB interface {
+	sqlx.ExtContext
+	sqlx.PreparerContext
+
+	PrepareNamedContext(ctx context.Context, query string) (*sqlx.NamedStmt, error)
+}
```
Al2Klimov, line 69: Also the whole thing seems just a subset of DB&Tx. I personally would unexport and move it...

database/db.go
```diff
@@ -652,6 +652,30 @@ func (db *DB) Delete(
...
+
+	if err = fn(tx); err != nil {
+		return err
+	}
+
+	if err = tx.Commit(); err != nil {
+		return errors.Wrap(err, "DB.RunInTx: cannot commit a database transaction")
+	}
+
+	return nil
```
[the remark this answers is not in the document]
yhabteab: Oh cool! I wasn't aware you can even just wrap nil error and would still return nil.

database/contracts.go
yhabteab: Wouldn't you also want to use this interface type elsewhere where a `DB` or `Tx` is required?

database/utils.go
```diff
@@ -43,6 +45,49 @@ func SplitOnDupId[T IDer]() com.BulkChunkSplitPolicy[T] {
...
+		// closed manually later on and the error is handled gracefully (if any).
+		defer func() { _ = ps.Close() }()
+
+		if err = ps.GetContext(ctx, &resultID, arg); err != nil {
+			return types.Int{}, CantPerformQuery(err, query)
+		}
+
+		if err = ps.Close(); err != nil {
+			return types.Int{}, errors.Wrapf(err, "cannot close prepared statement %q", query)
+		}
```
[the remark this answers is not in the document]
yhabteab: How should I understand your question, you are asking me why I am closing the prepared statement?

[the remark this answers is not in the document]
Al2Klimov: Why are you closing it _here_ given you're also closing it in a defer?

database/utils.go
```diff
@@ -43,6 +44,47 @@ func SplitOnDupId[T IDer]() com.BulkChunkSplitPolicy[T] {
...
+func InsertObtainID(ctx context.Context, conn TxOrDB, stmt string, arg any) (int64, error) {
+	var resultID int64
+	switch conn.DriverName() {
+	case PostgreSQL:
+		query := stmt + " RETURNING id"
+		ps, err := conn.PrepareNamedContext(ctx, query)
+		if err != nil {
+			return 0, errors.Wrapf(err, "cannot prepare %q", query)
+		}
+		// We're deferring the ps#Close invocation here just to be on the safe side, otherwise it's
```
Al2Klimov, line 62:
```suggestion
		// We're deferring the ps.Close invocation here just to be on the safe side, otherwise it's
```

yhabteab: Why?

Al2Klimov: \# is for an unspecified instance of type ps. But ps IS a specific instance – of type statement.

database/contracts.go
Al2Klimov: For me the last insert ID makes only sense for immediate subsequent processing – which just screams _transaction_. So I'd just allow Tx as parameter for now unless/until someone finds an actual use case with DB. (But then I'd:

1. unexport the interface OR
2. aggregate there all interfaces and non-interface methods both Tx and DB implement.)

database/utils.go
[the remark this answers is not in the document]
Al2Klimov: The error you're returning here isn't even necessary as the operation this function is for already completed successfully at this point.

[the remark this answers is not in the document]
yhabteab: Isn't that the same as what **YOU** are doing here?
https://github.com/Icinga/icinga-go-library/blob/2d47d95fe9d8d3d43d359de8569d9abdf6fecb9d/database/utils.go#L65-L74

database/utils.go
Al2Klimov: And now this two-case switch could be an _if_.

database/utils.go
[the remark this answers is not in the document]
Al2Klimov: Wait, do you really have to use a prepared statement in the first place? Isn't there a direct GetContext()?

database/utils.go
yhabteab: And why is that necessary? I think you are just focussing on the non-essential things.

lippserd (changes requested):
Please split pr the into two.

yhabteab:
The commits are split into their own PRs so that this is no longer necessary.

PR #19, opened 2024-05-27
workflows: Fix `gosec` errors & enable code annotations

yhabteab:
Enables code annotations in case of failures and fixes the actions triggered by the main branch.
https://github.com/Icinga/icinga-go-library/actions/runs/7783022809/job/21220618068

This PR suppresses also some of the reported existing `gosec` errors and solves some of them as well and drops the `only-new-issue` flag set for the go linter.

```bash
$ gosec ./...
---
Results:

Summary:
  Gosec  : dev
  Files  : 40
  Lines  : 4269
  Nosec  : 4
  Issues : 0
```

### DIFF
--- a/backoff/backoff.go
+++ b/backoff/backoff.go
@@ -39,5 +39,5 @@ func jitter(n int64) int64 {
 		return 0
 	}
 
-	return n/2 + rand.Int63n(n/2)
+	return n/2 + rand.Int63n(n/2) // #nosec G404 -- Use of weak random number generator - we don't need crypto/rand here though.
 }
--- a/config/config.go
+++ b/config/config.go
@@ -25,6 +25,7 @@ func FromYAMLFile(name string, v Validator) error {
 		return errors.Wrapf(ErrInvalidArgument, "non-nil pointer expected, got %T", v)
 	}
 
+	// #nosec G304 -- Potential file inclusion via variable - Its purpose is to load any file name that is passed to it, so doesn't need to validate anything.
 	f, err := os.Open(name)
 	if err != nil {
 		return errors.Wrap(err, "can't open YAML file "+name)
--- a/config/tls.go
+++ b/config/tls.go
@@ -22,7 +22,7 @@ func (t *TLS) MakeConfig(serverName string) (*tls.Config, error) {
 		return nil, nil
 	}
 
-	tlsConfig := &tls.Config{}
+	tlsConfig := &tls.Config{MinVersion: tls.VersionTLS12}
 	if t.Cert == "" {
 		if t.Key != "" {
 			return nil, errors.New("private key given, but client certificate missing")
--- a/version/version.go
+++ b/version/version.go
@@ -124,7 +124,7 @@ func (o *osRelease) DisplayVersion() string {
 // readOsRelease reads and parses the os-release file.
 func readOsRelease() (*osRelease, error) {
 	for _, path := range []string{"/etc/os-release", "/usr/lib/os-release"} {
-		f, err := os.Open(path)
+		f, err := os.Open(path) // #nosec G304 -- Potential file inclusion via variable - Hard-coded files, so not affected by this issue.
 		if err != nil {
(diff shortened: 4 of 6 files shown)


yhabteab:
refs https://github.com/Icinga/icinga-go-library/pull/9

config/tls.go
```diff
@@ -22,7 +22,7 @@ func (t *TLS) MakeConfig(serverName string) (*tls.Config, error) {
 		return nil, nil
 	}
 
-	tlsConfig := &tls.Config{}
+	tlsConfig := &tls.Config{MinVersion: tls.VersionTLS12}
```
yhabteab, line 25: Go redis does the same thing here too.
https://github.com/redis/go-redis/blob/master/options.go#L314-L317

config/config.go
```diff
@@ -25,7 +26,7 @@ func FromYAMLFile(name string, v Validator) error {
 		return errors.Wrapf(ErrInvalidArgument, "non-nil pointer expected, got %T", v)
 	}
 
-	f, err := os.Open(name)
+	f, err := os.Open(filepath.Clean(name))
```
oxzi, line 29: Okay, but how or why does this please the linter?

Based on [filepath.Cleans documentation](https://pkg.go.dev/path/filepath#Clean) I cannot see what security this might bring? This should only sanitize the path, but shouldn't change the target.

backoff/backoff.go
```diff
@@ -39,5 +39,5 @@ func jitter(n int64) int64 {
 		return 0
 	}
 
-	return n/2 + rand.Int63n(n/2)
```
oxzi, line 42: Even when `math.rand` and `crypto/rand` are [getting closer together](https://go.dev/blog/chacha8rand), we could also just use `crypto/rand` here.

version/version.go
```diff
@@ -124,7 +125,7 @@ func (o *osRelease) DisplayVersion() string {
 // readOsRelease reads and parses the os-release file.
 func readOsRelease() (*osRelease, error) {
 	for _, path := range []string{"/etc/os-release", "/usr/lib/os-release"} {
-		f, err := os.Open(path)
+		f, err := os.Open(filepath.Clean(path))
```
oxzi, line 128: Same as above, but even more obvious. What should be cleaned here? Those strings are hard-coded absolute paths. I would further ignore this linter warning as adding `filepath.Clean` just obscures things.

config/config.go
yhabteab: The `gosec` reports this as an error and causes the GitHub actions to fail, so I had to either fix it or suppress it.
```bash
[/Users/yhabteab/Workspace/go/icinga-go-library/config/config.go:28] - G304 (CWE-22): Potential file inclusion via variable (Confidence: HIGH, Severity: MEDIUM)
    27: 
  > 28:         f, err := os.Open(name)
    29:         if err != nil {
```

backoff/backoff.go
yhabteab: It needs way more code changes with `crypto/rand` than just replacing it like `error` handling, `bigInt` usage etc, and since we don't actually need cryptographically secure randomness here, I chose to just suppress it.

version/version.go
yhabteab: Same here as well!

config/config.go
yhabteab: It's complaining about this issue: https://securego.io/docs/rules/g304.html, but it doesn't seem to get a bit more context in which such an issue could be dangerous.

oxzi: Yes, but just adding a `filepath.Clean` call does not solve the addressed issue. This would also require the prefix check as demonstrated in the "The right way" code listing on your URL. Thus, this would have no effect except confusing the reader and satisfying an incomplete linter check.

Furthermore, the path should be a "safe" one as it points to the YAML to be parsed. There is no reference location to be compared against. Thus, I would say ignore this linter warning as it makes no sense in this context.

oxzi (approved):

PR #20, opened 2024-05-27
MySQL/MariaDB set session vars unittests

yhabteab:
refs #10

### DIFF
--- a/.github/workflows/sql.yml
+++ b/.github/workflows/sql.yml
@@ -1,0 +1,69 @@
+name: SQL
+
+on:
+  push:
+    branches:
+      - main
+  pull_request: {}
+
+jobs:
+  mysql:
+    name: ${{ matrix.database.name }}
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        database:
+          - {name: MySQL 5.7,      image: "mysql:5.7"}
+          - {name: MySQL 8.0,      image: "mysql:8.0"}
+          - {name: MySQL latest,   image: "mysql:latest"}
+          - {name: MariaDB 10.1,   image: "mariadb:10.1"}
+          - {name: MariaDB 10.2,   image: "mariadb:10.2"}
+          - {name: MariaDB 10.3,   image: "mariadb:10.3"}
+          - {name: MariaDB 10.4,   image: "mariadb:10.4"}
+          - {name: MariaDB 10.5,   image: "mariadb:10.5"}
+          - {name: MariaDB 10.6,   image: "mariadb:10.6"}
+          - {name: MariaDB 10.7,   image: "mariadb:10.7"}
+          - {name: MariaDB 10.11,  image: "mariadb:10.11"}
+          - {name: MariaDB 11.0,   image: "mariadb:11.0"}
+          - {name: MariaDB latest, image: "mariadb:latest"}
+
+    env:
+      ICINGAGOLIBRARY_TESTS_DB_TYPE: mysql
+      ICINGAGOLIBRARY_TESTS_DB: icinga_unittest
+      ICINGAGOLIBRARY_TESTS_DB_USER: root
+      ICINGAGOLIBRARY_TESTS_DB_PASSWORD: password
+      ICINGAGOLIBRARY_TESTS_DB_HOST: 127.0.0.1
+      ICINGAGOLIBRARY_TESTS_DB_PORT: 3306
+
+    services:
+      mysql:
+        image: ${{ matrix.database.image }}
+        env:
+          MYSQL_ROOT_PASSWORD: ${{ env.ICINGAGOLIBRARY_TESTS_DB_PASSWORD }}
+          MYSQL_DATABASE: ${{ env.ICINGAGOLIBRARY_TESTS_DB }}
+        # Wait for the containers to become ready
+        options: >-
+          --health-cmd "${{ (startsWith(matrix.database.image, 'mysql:') || startsWith(matrix.database.image, 'mariadb:10')) && 'mysqladmin ping' || 'healthcheck.sh --connect --innodb_initialized' }}"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 10
+        ports:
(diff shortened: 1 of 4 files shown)


.github/workflows/sql.yml
```diff
@@ -0,0 +1,65 @@
...
+  mysql:
+    name: ${{ matrix.database.name }}
+    runs-on: ubuntu-latest
+
+    strategy:
+      fail-fast: false
+      matrix:
+        database:
+          - {name: MySQL 5.7,      image: "mysql:5.7"}
+          - {name: MySQL 8.0,      image: "mysql:8.0"}
```
Al2Klimov, line 19: Are these two really all MySQL versions we support?

.github/workflows/sql.yml
```diff
@@ -0,0 +1,71 @@
...
+      mysql:
+        image: ${{ matrix.database.image }}
+        env:
+          MYSQL_ROOT_PASSWORD: password
+          MYSQL_DATABASE: icinga_unittest
+          MARIADB_ROOT_PASSWORD: password # mariadb >= 11
+          MARIADB_DATABASE: icinga_unittest # mariadb >= 11
+        # Wait for the containers to become ready
+        options: >-
+          --health-cmd ${{ matrix.database.compat != 'false' && '"mysqladmin ping"' || '"healthcheck.sh --connect --innodb_initialized"' }}
```
julianbrost, line 50: Would something like the following also work, i.e. provide a default `--health-cmd` and allow to override it the matrix?

```suggestion
          --health-cmd "${{ matrix.database.healthcmd || 'mysqladmin ping' }}"
```

yhabteab: Maybe I didn't test it, but why? That way you would have to define `healthcmd` for every MariaDB version that is not compatible with MySQL.

julianbrost: It's just that `compat: 'false'` in the matrix definition is pretty meaningless. My suggestion would be to try to make it more expressive.

yhabteab: > It's just that compat: 'false' in the matrix definition is pretty meaningless.

How is that "meaningless"? It just indicates whether that specific mariadb image is fully mysql-compatible or not, and I don't consider that as "meaningless" at all. If you like, I can add a few sentences as a comment but in any case, I find having `compat: 'false'` in the matrix far better than defining the entire health command every time.

.github/workflows/sql.yml
```diff
@@ -0,0 +1,71 @@
...
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 10
```
Al2Klimov, line 53: * Huh? What happened to #21?

yhabteab: Sorry, reverted!

Al2Klimov: Didn't it work?

.github/workflows/sql.yml
```diff
@@ -0,0 +1,71 @@
...
+      ICINGAGOLIBRARY_TESTS_DB_PORT: 3306
+
+    services:
+      mysql:
+        image: ${{ matrix.database.image }}
+        env:
+          MYSQL_ROOT_PASSWORD: password
+          MYSQL_DATABASE: icinga_unittest
+          MARIADB_ROOT_PASSWORD: password # mariadb >= 11
+          MARIADB_DATABASE: icinga_unittest # mariadb >= 11
```
Al2Klimov, line 47: Sure this is necessary? https://github.com/Icinga/icinga-notifications/actions/runs/9972045687/job/27554443566

.github/workflows/sql.yml
yhabteab: There is also `mysql:latest`, so which version do you want to add?

database/testutils.go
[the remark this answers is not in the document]
yhabteab: I just wanted it to be importable by other projects as well, but have now reverted it.

.github/workflows/sql.yml
yhabteab: Actually, it seems not to be necessary!

database/testutils.go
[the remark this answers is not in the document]
julianbrost: That can be done, but then the function should be in a separate package to prevent all users of the `database` package from pulling in test-only dependencies like `stretchr/testify`.

database/set_mysql_session_vars_test.go
julianbrost, line 1: Typically, the test files are named after the files that contain the code they test. So this should be `utils_test.go` as it tests something from `utils.go`.

database/utils.go
```diff
@@ -44,22 +44,22 @@ func SplitOnDupId[T IDer]() com.BulkChunkSplitPolicy[T] {
...
-	galeraOpts := "SET SESSION wsrep_sync_wait=" + strconv.FormatInt(wsrepSyncWait, 10)
-
-	_, err := conn.(driver.ExecerContext).ExecContext(ctx, galeraOpts, nil)
+// NOTE: It is unsafe to use this function with untrusted/user supplied inputs and poses an SQL injection,
+// because it doesn't use a prepared statement, but executes the SQL command directly with the provided inputs.
+//
+// When the "SET SESSION" command fails with "Unknown system variable (1193)", the error will be silently
+// dropped but returns all other database errors.
+func unsafeSetSessionVariableIfExists(ctx context.Context, conn driver.Conn, variable string, value any) error {
+	stmt := fmt.Sprintf("SET SESSION %s=%v", variable, value)
```
julianbrost, line 55: I would change the type of `value` to `string` to make the behavior of this function a bit more obvious. With `any`, one might expect it to use the [`driver.Valuer`](https://pkg.go.dev/database/sql/driver#Valuer) interface, i.e. that it would work with something like `types.String` which it doesn't. That change also wouldn't make the function harder to use, you'd have to pass `fmt.Sprint(x)` instead of `x` at worst.

julianbrost (changes requested):

julianbrost (changes requested):

julianbrost (approved):

PR #21, opened 2024-05-28
GHA: SQL: don't override health checks

Al2Klimov:
GHA should work now!


yhabteab:
OMG 😦! It's that easy?

PR #22, opened 2024-05-29
Restore deleted `DB#BuildColumns()`

yhabteab:
Simply restores the deleted method `DB#BuildColumns()`, as there is no other direct way to generate columns, since the column mapper is hidden.


oxzi (approved):
For reference, the method was removed within 102481adcdfcc54583eaf8627dad4d2be818df18.

The implementation looks fine, especially with the cloned `ColumnMap` to not mess with the internal cache.

PR #23, opened 2024-06-27
redis.Client#XReadUntilResult(): also re-try timeout errors

Al2Klimov:
We need to retry the XREAD commands in the following situations:
    
- If Go Redis returns redis.Nil, it means no data was read from Redis — e.g. when the keys don’t exist yet, and we will need to retry the operation again.
- To prevent surpassing Go Redis's internal maximum retries or any other I/O timeouts [^1], it's important to set a block timeout greater than zero for the XREAD commands, see the `a.Block` assignment. However, setting a block timeout means that Go Redis will not retry any errors internally and will instead return an I/O timeout error when exceeding the timeout. Thus, we need to handle this here and retry it again. On the other hand, an I/O timeout could also mean a `context.DeadlineExceeded` error, which is not retryable, so we have to check for context termination by ourselves via `ctx.Err()`.

fixes https://github.com/Icinga/icingadb/issues/786
ref/IP/54215

[^1]: https://github.com/redis/go-redis/issues/2131

### DIFF
--- a/redis/client.go
+++ b/redis/client.go
@@ -216,7 +216,19 @@ func (c *Client) XReadUntilResult(ctx context.Context, a *redis.XReadArgs) ([]re
 		cmd := c.XRead(ctx, a)
 		streams, err := cmd.Result()
 		if err != nil {
-			if errors.Is(err, redis.Nil) {
+			// We need to retry the XREAD commands in the following situations:
+			// - If Go Redis returns redis.Nil, it means no data was read from Redis — e.g. when the keys don’t
+			//  exist yet, and we will need to retry the operation again.
+			//
+			// - To prevent surpassing Go Redis's internal maximum retries or any other I/O timeouts [^1], it's
+			//  important to set a block timeout greater than zero for the XREAD commands, see the "a.Block" above.
+			//  However, setting a block timeout means that Go Redis will not retry any errors internally and will
+			//  instead return an I/O timeout error when exceeding the timeout. Thus, we need to handle this here and
+			//  retry it again. On the other hand, an I/O timeout could also mean a context.DeadlineExceeded error,
+			//  which is not retryable, so we have to check for context termination by ourselves via ctx.Err().
+			//
+			// [^1]: https://github.com/redis/go-redis/issues/2131
+			if (errors.Is(err, redis.Nil) || retry.Retryable(err)) && ctx.Err() == nil {
 				continue
 			}
 


Al2Klimov:
CC @julianbrost @tbauriedel

yhabteab (commented):
This is not the right way of resolving the underlying issue. I mean, retrying the error might actually workaround it, but that's not the proper solution. [`go-redis`](https://github.com/redis/go-redis/blob/v9.5.1/redis.go#L417-L439) is capable of retrying i/o timeouts, but this code path is [only effective](https://github.com/redis/go-redis/blob/v9.5.1/redis.go#L427-L428) if you do not explicitly define a read timeout, which is [not the case here](https://github.com/Icinga/icingadb/blob/v1.2.0/pkg/icingaredis/client.go#L194-L197).
https://github.com/redis/go-redis/blob/v9.5.1/stream_commands.go#L163-L165

And our default [`BlockTimeout`](https://github.com/Icinga/icingadb/blob/v1.2.0/pkg/icingaredis/client.go#L31) happens to be `1s`, which effectively disables the go-redis internal retry handling and overrides its [read timeout](https://github.com/redis/go-redis/blob/v9.5.1/redis.go#L449-L456) of `10s`. As I mentioned above, retrying it might fix the issue, but we should consider dropping the default `BlockTimeout` and only setting it if explicitly configured by the user.

Al2Klimov:
... so the user could run into ref/IP/54215 depending on their config? Scary IMAO. In contrast, my implementation works always.

yhabteab:
> Scary IMAO. In contrast, my implementation works always.

No, it is not! If you as a user configure `30s` as a timeout, then you should not be surprised if Icinga DB gives up after exactly these seconds and thus it is not something that Icinga DB should keep retrying.

julianbrost:
> As I mentioned above, retrying it might fix the issue, but we should consider dropping the default `BlockTimeout` and only setting it if explicitly configured by the user.

That block in the XREAD is set for a reason, see cd96f0de6f33b6750ba44d3934d4380a62ab14a6. So if you'd just undo that, that problem would probably come back.

> ... so the user could run into ref/IP/54215 depending on their config?

Please don't just reference internal tickets as part of your argument without any kind of summary. It's annoying to look up the whole ticket to get what you're trying to say and for everyone outside of Icinga, it's even impossible to follow.

redis/client.go
```diff
@@ -220,6 +220,13 @@ func (c *Client) XReadUntilResult(ctx context.Context, a *redis.XReadArgs) ([]re
 				continue
 			}
 
+			var timeout interface {
+				Timeout() bool
+			}
+			if errors.As(err, &timeout) && timeout.Timeout() && !errors.Is(err, context.DeadlineExceeded) {
+				continue
+			}
```
julianbrost, line 228: 1. We already have a function with the purpose to test if an error is considered retryable, why isn't it used here, is there a reason why some of the errors it consideres retryable shouldn't be retried here? https://github.com/Icinga/icinga-go-library/blob/420fbff2ee6fc4ff2bc53df3fe9d4f1ffb64570c/retry/retry.go#L135-L201
1. Explicitly checking for `context.DeadlineExceeded` feels like a very narrow fix for https://github.com/Icinga/icinga-go-library/pull/23#discussion_r1705689655. What about having `ctx.Err() == nil` as an additional condition for a retry? That sounds like a sane check in general, avoids the problem as well, and doesn't rely on any context error internals.

So maybe just do something like this instead (I can't mark it as a suggestion as it also subsumes the previous `if` which is one line out of the diff context):

```go
			if (errors.Is(err, redis.Nil) || retry.Retryable(err)) && ctx.Err() == nil {
				continue
			}
```

julianbrost (approved):
After building Icinga DB pinned to this branch of Icinga Go Library, the behavior in the following test scenario changes:

```sh
killall -STOP redis-server && date && sleep 3m && killall -CONT redis-server && date
```

### Before

Redis was paused from 16:44:59Z to 16:47:59Z:

```
2024-08-09T15:45:10.262Z    FATAL   icingadb    read tcp 172.18.0.21:59398->172.18.0.5:6379: i/o timeout
can't perform "[xread count 4096 block 1000 streams icinga:history:stream:notification 1723218296379-1]"
github.com/icinga/icinga-go-library/redis.WrapCmdErr
    /go/pkg/mod/github.com/icinga/icinga-go-library@v0.3.1/redis/utils.go:15
github.com/icinga/icinga-go-library/redis.(*Client).XReadUntilResult
    /go/pkg/mod/github.com/icinga/icinga-go-library@v0.3.1/redis/client.go:222
github.com/icinga/icingadb/pkg/icingadb/history.Sync.readFromRedis
    /icingadb-src/pkg/icingadb/history/sync.go:112
github.com/icinga/icingadb/pkg/icingadb/history.Sync.Sync.func1
    /icingadb-src/pkg/icingadb/history/sync.go:81
golang.org/x/sync/errgroup.(*Group).Go.func1
    /go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
    /usr/local/go/src/runtime/asm_amd64.s:1695
can't read history
github.com/icinga/icingadb/pkg/icingadb/history.Sync.readFromRedis
    /icingadb-src/pkg/icingadb/history/sync.go:114
github.com/icinga/icingadb/pkg/icingadb/history.Sync.Sync.func1
    /icingadb-src/pkg/icingadb/history/sync.go:81
golang.org/x/sync/errgroup.(*Group).Go.func1
    /go/pkg/mod/golang.org/x/sync@v0.7.0/errgroup/errgroup.go:78
runtime.goexit
    /usr/local/go/src/runtime/asm_amd64.s:1695
```

### After

Redis was paused from 16:59:21Z to 17:02:21Z:

```
2024-08-09T16:00:20.141Z    WARN    heartbeat   Lost Icinga heartbeat   {"timeout": "1m0s"}
2024-08-09T16:00:20.141Z    ERROR   high-availability   Lost heartbeat  
2024-08-09T16:01:20.141Z    WARN    heartbeat   Waiting for Icinga heartbeat
2024-08-09T16:02:20.141Z    WARN    heartbeat   Waiting for Icinga heartbeat
2024-08-09T16:02:21.735Z    INFO    heartbeat   Received Icinga heartbeat   {"environment": "6cc4013ece4ffdbb275a4a763460507d17203146"}
2024-08-09T16:02:21.736Z    ERROR   high-availability   Received heartbeat from the past    {"time": "2024-08-09T15:59:24.555Z"}
2024-08-09T16:02:21.739Z    ERROR   high-availability   Received heartbeat from the past    {"time": "2024-08-09T15:59:25.571Z"}
2024-08-09T16:02:23.170Z    INFO    high-availability   Another instance is active  {"instance_id": "fe568d4d2a074dea822696dcd79801df", "environment": "6cc4013ece4ffdbb275a4a763460507d17203146", "heartbeat": "2024-08-09T16:02:23.030Z", "heartbeat_age": "140.936828ms"} 
```

It now survives. Logging could be better though, but this probably has to be handled on the Icinga DB side.

redis/client.go
```diff
@@ -216,7 +216,7 @@ func (c *Client) XReadUntilResult(ctx context.Context, a *redis.XReadArgs) ([]re
 		cmd := c.XRead(ctx, a)
 		streams, err := cmd.Result()
 		if err != nil {
-			if errors.Is(err, redis.Nil) {
+			if (errors.Is(err, redis.Nil) || retry.Retryable(err)) && ctx.Err() == nil {
```
oxzi, line 219: Coming back to this diff after just one week, I have totally forgotten why there is a `ctx.Err() == nil`. Could you please add a short comment with the reasoning from <https://github.com/Icinga/icinga-go-library/pull/23#discussion_r1705689655> and <https://github.com/Icinga/icinga-go-library/pull/23#discussion_r1707154827>. Otherwise, LGTM. Thanks.

Al2Klimov: IMAO it's not limited to those comments. One could also pass a context and cancel it, regardless the current error. Then, it makes no sense to re-try.

But yes, `ctx.Err()==nil` is not our idiomatic way of checking whether ctx hasn't been e.g. cancelled, yet. What about instead code which doesn't need such commenting in the first place, i.e. select on ctx.Done() as usual?

julianbrost: I can't follow why the following (i.e. the current version of the PR) wouldn't need an explaining comment when checking `ctx.Err() == nil` does:

https://github.com/Icinga/icinga-go-library/blob/2efd1f7f182912edaf40928a68a58b7d07a0d224/redis/client.go#L220-L224

It's pretty obvious to me that both does the same, the `select` is just more bloated.

> But yes, `ctx.Err()==nil` is not our idiomatic way of checking whether ctx hasn't been e.g. cancelled, yet. What about instead code which doesn't need such commenting in the first place, i.e. select on ctx.Done() as usual?

Is it idiomatic? It's simply necessary if you want to use a context to unblock waiting send/receive on a channel.

yhabteab: @Al2Klimov can you please revert this to the previous version and just add a simple comment as to why it is necessary to check for `ctx.Error()`. Introducing a `select` case just to check if a ctx has an error instead of just adding a simple comment is ridiculous and keeps the PR open for nothing.

yhabteab (dismissed):
Tested with: `docker kill -s STOP redis && date && sleep 180 && docker kill -s CONT redis && date`

**Before:**
```bash
$ docker kill -s STOP redis && date && sleep 180 && docker kill -s CONT redis && date
redis
Mo 16 Sep 2024 15:24:02 CEST

--------
2024-09-16T15:24:13.040+0200    FATAL   main    read tcp 127.0.0.1:51709->127.0.0.1:6380: i/o timeout
can't perform "[xread block 1000 streams icinga:schema 1726492901357-0]"
github.com/icinga/icinga-go-library/redis.WrapCmdErr
        /Users/yhabteab/Workspace/go/icinga-go-library/redis/utils.go:15
github.com/icinga/icinga-go-library/redis.(*Client).XReadUntilResult
        /Users/yhabteab/Workspace/go/icinga-go-library/redis/client.go:223
main.checkRedisSchema
        /Users/yhabteab/Workspace/go/icingadb/cmd/icingadb/main.go:392
main.monitorRedisSchema
        /Users/yhabteab/Workspace/go/icingadb/cmd/icingadb/main.go:374
runtime.goexit
        /opt/homebrew/opt/go/libexec/src/runtime/asm_arm64.s:1222
can't read Redis schema version
main.checkRedisSchema
        /Users/yhabteab/Workspace/go/icingadb/cmd/icingadb/main.go:396
main.monitorRedisSchema
        /Users/yhabteab/Workspace/go/icingadb/cmd/icingadb/main.go:374
runtime.goexit
        /opt/homebrew/opt/go/libexec/src/runtime/asm_arm64.s:1222
exit status 1
```

**After:**
```bash
2024-09-16T15:19:39.225+0200    ERROR   high-availability       Lost heartbeat
2024-09-16T15:19:39.225+0200    WARN    main    Handing over    {"reason": "lost heartbeat"}
2024-09-16T15:20:39.226+0200    WARN    heartbeat       Waiting for Icinga heartbeat
2024-09-16T15:21:39.227+0200    WARN    heartbeat       Waiting for Icinga heartbeat
2024-09-16T15:21:41.002+0200    INFO    heartbeat       Received Icinga heartbeat       {"environment": "e9698ca8a2cd336a55a948ea51903878d13869bd"}
2024-09-16T15:21:41.002+0200    ERROR   high-availability       Received heartbeat from the past        {"time": "2024-09-16T15:18:59.371+0200"}
2024-09-16T15:21:41.371+0200    INFO    main    Taking over     {"reason": "no other instance is active"}
2024-09-16T15:21:41.376+0200    INFO    main    Starting config sync
2024-09-16T15:21:41.376+0200    INFO    main    Starting initial state sync
2024-09-16T15:21:41.376+0200    INFO    main    Starting overdue sync
```

redis/client.go
```diff
@@ -216,7 +216,9 @@ func (c *Client) XReadUntilResult(ctx context.Context, a *redis.XReadArgs) ([]re
 		cmd := c.XRead(ctx, a)
 		streams, err := cmd.Result()
 		if err != nil {
-			if errors.Is(err, redis.Nil) {
+			// context.DeadlineExceeded implements both Temporary() and Timeout() which makes it retryable.
```
lippserd, line 219: I had to read through the PR and issue and all the comments to understand this change. Please adjust the comment accordingly:

- Why we need to explicitly retry here.
- Why the `ctx.Err() == nil` check is necessary.

Please also adjust the commit description with proper details for this change. The commit message is misleading as well.

lippserd (changes requested):

lippserd (approved):
Thanks for the adjustments @yhabteab.

PR #24, opened 2024-07-04
Test types.String

Al2Klimov:


types/string_test.go
```diff
@@ -0,0 +1,126 @@
...
+	}{
+		{"empty", ""},
+		{"nul", "\x00"},
+		{"space", " "},
+		{"multiple", "abc"},
+	}
+
+	for _, st := range subtests {
+		t.Run(st.name, func(t *testing.T) {
+			require.Equal(t, String{NullString: sql.NullString{String: st.io, Valid: true}}, MakeString(st.io))
```
oxzi, line 23: What's the benefit of this test function? It compares if a "manually" created `String` is equal to the exact same code being executed within the `MakeString` call.

oxzi (approved):

types/string_test.go
julianbrost: Test coverage 🤪

The most useful thing about this test case probably is that it checks that `MakeString("")` doesn't result in SQL `NULL`.

PR #32, opened 2024-07-05
Test utils/

Al2Klimov:


utils/utils_test.go
```diff
@@ -1,10 +1,195 @@
...
+	}
+
+	for _, st := range subtests {
+		t.Run(st.name, func(t *testing.T) {
+			require.Equal(t, st.output, Ellipsize(st.s, st.limit))
+		})
+	}
+}
+
+func TestMaxInt(t *testing.T) {
```
oxzi, line 134: I would advise against testing this helper function. Furthermore, I would argue for removing it, as it has landed in the Go standard library a while ago with 1.21.0 as [`builtin.max`](https://pkg.go.dev/builtin#max).

utils/utils_test.go
```diff
@@ -1,10 +1,195 @@
...
+					require.True(t, ok, "receiving should return a value")
+					require.Equal(t, expected, actual)
+				case <-time.After(10 * time.Millisecond):
+					require.Fail(t, "receiving should not block")
+				}
+			}
+
+			select {
+			case _, ok := <-batches:
+				require.False(t, ok, "receiving from channel should not return anything")
```
oxzi, line 49: Eases debugging:

```suggestion
			case out, ok := <-batches:
				require.False(t, ok, "receiving from channel should not return anything, received %q", out)
```

utils/utils_test.go
```diff
@@ -1,10 +1,195 @@
...
+		{"a", []string{"a"}, 1, [][]string{{"a"}}},
+		{"a2", []string{"a"}, 2, [][]string{{"a"}}},
+		{"a_b", []string{"a", "b"}, 1, [][]string{{"a"}, {"b"}}},
+		{"ab", []string{"a", "b"}, 2, [][]string{{"a", "b"}}},
+		{"ab3", []string{"a", "b"}, 3, [][]string{{"a", "b"}}},
+		{"a_b_c", []string{"a", "b", "c"}, 1, [][]string{{"a"}, {"b"}, {"c"}}},
+		{"ab_c", []string{"a", "b", "c"}, 2, [][]string{{"a", "b"}, {"c"}}},
+		{"abc", []string{"a", "b", "c"}, 3, [][]string{{"a", "b", "c"}}},
+		{"abc4", []string{"a", "b", "c"}, 4, [][]string{{"a", "b", "c"}}},
+	}
```
oxzi, line 30: As the `count` parameter is a signed integer, try negative numbers :smirk: 

I would suggest making the function documentation more obvious and stating that `count <= 0` will either panic or be rewritten to 1. A `count` of 0 returns an infinite loop emitting empty string arrays until eternity and a negative `count` results in a slice bounds out of range access.

```go
{"abc0-to-infinity-and-beyond", []string{"a", "b", "c"}, 0, [][]string{{}, {}, {}}},
{"abc-1-crash", []string{"a", "b", "c"}, -1, nil},
```

```
=== RUN   TestBatchSliceOfStrings/abc0-to-infinity-and-beyond
    utils_test.go:51: 
        	Error Trace:	/home/REDACTED/git/github.com/Icinga/icinga-go-library/utils/utils_test.go:51
        	Error:      	Should be false
        	Test:       	TestBatchSliceOfStrings/abc0-to-infinity-and-beyond
        	Messages:   	receiving from channel should not return anything, received []
=== RUN   TestBatchSliceOfStrings/abc-1-crash
panic: runtime error: slice bounds out of range [:-1]
```

utils/utils_test.go
```diff
@@ -1,10 +1,195 @@
...
+func TestChecksum(t *testing.T) {
+	subtests := []struct {
+		name  string
+		input any
+	}{
+		{"empty_string", ""},
+		{"empty_bytes", []byte(nil)},
+		{"space_string", " "},
+		{"space_bytes", []byte(" ")},
+	}
```
oxzi, line 66: The `Checksum` function panics for other `input` types. Please test those as well.

utils/utils_test.go
```diff
@@ -1,10 +1,195 @@
...
+				require.Fail(t, "receiving should not block")
+			}
+		})
+	}
+}
+
+func TestChecksum(t *testing.T) {
+	subtests := []struct {
+		name  string
+		input any
```
oxzi, line 60: Why don't also check the output? Doing so ensures that nobody refactors the function to break the checksum algorithm.

oxzi (changes requested):

utils/utils_test.go
oxzi: #65

utils/utils_test.go
Al2Klimov: Where I just ref an issue/PR (like you https://github.com/Icinga/icinga-go-library/pull/32#discussion_r1718314038), I prefix it with * and a space, so I show its title and status. E.g.:

* #66

oxzi: After this got merged, please include it in the tests, e.g., by adding another `shouldPanic` field to the table test struct.

oxzi (changes requested):
There are still two open review comments from my previous review: <https://github.com/Icinga/icinga-go-library/pull/32#discussion_r1778215902> and <https://github.com/Icinga/icinga-go-library/pull/32#discussion_r1717149581>. Please address those and feel free to re-request another review again.

utils/utils_test.go
```diff
@@ -1,10 +1,201 @@
...
+				require.False(t, ok, "receiving from channel should not return anything")
+			case <-time.After(10 * time.Millisecond):
+				require.Fail(t, "receiving should not block")
+			}
+		})
+	}
+
+	for _, i := range []int{0, -1, -2, -30} {
+		t.Run(fmt.Sprint(i), func(t *testing.T) {
+			require.Panics(t, func() { BatchSliceOfStrings(context.Background(), nil, i) })
```
Al2Klimov, line 60: Here, I assert panic on negative numbers.

utils/utils_test.go
```diff
@@ -1,10 +1,201 @@
...
+		{"int", 0},
+		{"float", 0.0},
+		{"struct", struct{}{}},
+		{"slice", []string{}},
+		{"map", map[string]string{}},
+	}
+
+	for _, st := range unsupported {
+		t.Run(st.name, func(t *testing.T) {
+			require.Panics(t, func() { Checksum(st.input) })
```
Al2Klimov, line 98: And here, I assert panic on unsupported types.

oxzi:
This PR branch now has merge conflicts :crying_cat_face:

oxzi (approved):

PR #34, opened 2024-07-08
Test redis.Streams#Option()

Al2Klimov:


oxzi (approved):

redis/streams_test.go
```diff
@@ -0,0 +1,25 @@
...
+func TestStreams_Option(t *testing.T) {
+	subtests := []struct {
+		name   string
+		input  Streams
+		output []string
+	}{
+		{"nil", nil, []string{}},
+		{"empty", Streams{}, []string{}},
+		{"one", Streams{"key": "id"}, []string{"key", "id"}},
+		{"two", Streams{"key1": "id1", "key2": "id2"}, []string{"key1", "key2", "id1", "id2"}},
```
oxzi, line 17: This may fail. As we are dealing with a map-based type, the order cannot be predicted.

https://github.com/Icinga/icinga-go-library/blob/e7449283d69464b88dc5f47c2abf45049034ab8f/redis/streams.go#L4

This concrete subtest fails randomly for me. Please change your verification logic to check that the keys and ids are in the correct relative order.

oxzi (changes requested):

oxzi (approved):

oxzi (approved):
Retested again after rebasing against main.

PR #35, opened 2024-07-08
Test com.Atomic

Al2Klimov:


oxzi (approved):
Works as intended.

But honestly, I am unsure about the value of those tests as the tested code is a very thin wrapper around Go's `atomic.Value`, just adding generic type information.

lippserd:
I would rather remove that type completely. See Icinga/icingadb#812.

PR #36, opened 2024-07-08
Test com.Bulk()

Al2Klimov:


com/bulker_test.go
```diff
@@ -0,0 +1,114 @@
...
+									return
+								}
+							}
+
+							for _, v := range chunk {
+								if l.latency > 0 {
+									select {
+									case <-time.After(l.latency):
+									case <-ctx.Done():
+										return
```
oxzi, line 73: This is on indentation level ten, in a `go func` within two `t.Run` calls.

com/bulker_test.go
oxzi, line 1: The test are missing the multiple context done cases. Please add test cases with a context you are closing prematurely.

com/bulker_test.go
```diff
@@ -0,0 +1,129 @@
...
+						}
+					case <-time.After(time.Second):
+						require.Fail(t, "channel should not block")
+					}
+				})
+			}
+		})
+	}
+
+	t.Run("closed-ctx", func(t *testing.T) {
```
oxzi, line 115: Thanks for creating context-related tests. However, those are not really representative as they call `Bulk` with an already closed context. The more relevant cases are when the context gets cancelled half way through, e.g., the following lines:
- https://github.com/Icinga/icinga-go-library/blob/0cfa2158c09d222a017f74f5800af1d1c06c32e0/com/bulker.go#L72-L73
- https://github.com/Icinga/icinga-go-library/blob/0cfa2158c09d222a017f74f5800af1d1c06c32e0/com/bulker.go#L105-L106
- https://github.com/Icinga/icinga-go-library/blob/0cfa2158c09d222a017f74f5800af1d1c06c32e0/com/bulker.go#L152-L153

oxzi (approved):

oxzi (approved):
Retested again after rebasing against main.

PR #38, opened 2024-07-08
Test com/

Al2Klimov:


Al2Klimov:
Well, because it's so small.

lippserd:
> Well, because it's so small.

I don't understand why the code size should determine whether a function should be tested or not. Please add a test.

oxzi (approved):

PR #46, opened 2024-07-19
Test com.Cond

Al2Klimov:


com/cond_test.go
```diff
@@ -0,0 +1,80 @@
...
+	case <-time.After(time.Second / 10):
+	}
+
+	cond.Broadcast()
+
+	select {
+	case <-done:
+		require.Fail(t, "cond should still not be closed")
+	case <-cond.Done():
+		require.Fail(t, "cond should not be closed for round 2, yet")
```
oxzi, line 31: Is there a difference between those two cases? Both cases check the value of `cond.Done()`, returning `cond.done` which doesn't change over the time. Unless I'm missing something, this is the exact same channel.

com/cond_test.go
```diff
@@ -0,0 +1,80 @@
...
+		require.Fail(t, "cond should be ready")
+	}
+}
+
+func TestCond_Close(t *testing.T) {
+	cond := NewCond(context.Background())
+	done := cond.Done()
+	wait := cond.Wait()
+
+	require.NoError(t, cond.Close())
```
oxzi, line 52: In its current implementation, `Close()` always returns `nil`, even when the API doc states otherwise.

com/cond_test.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	select {
+	case _, ok := <-cond.Done():
+		if ok {
+			require.Fail(t, "new cond-closed channel should be closed")
+		}
+	case <-time.After(time.Second / 10):
+		require.Fail(t, "cond should be still closed")
+	}
```
oxzi, line 70: Again, same `.done` channel as already checked in the `select` above.

oxzi (changes requested):

com/cond_test.go
Al2Klimov: What you digged out is an implementation detail which tests don't rely on.

com/cond_test.go
Al2Klimov: What you digged out is an implementation detail which tests don't rely on.

oxzi (approved):

oxzi:
I have force-pushed a rebased version to trigger the CI. The prior runs were stalled for some reason.

PR #47, opened 2024-07-19
Test com.Counter

Al2Klimov:


com/counter_test.go
```diff
@@ -0,0 +1,32 @@
...
+	c.Add(23)
+	require.Equal(t, uint64(65), c.Val(), "unexpected new value")
+	require.Equal(t, uint64(65), c.Total(), "unexpected new total")
+}
+
+func TestCounter_Reset(t *testing.T) {
+	var c Counter
+
+	c.Add(42)
+	c.Reset()
```
oxzi, line 24: Why not check the return value?

```suggestion
	require.Equal(t, uint64(42), c.Reset(), "unexpected reset value")
```

oxzi (changes requested):
IMHO, the two test functions could be merged in one, as the only difference is that the second one has an additional `Reset()` call.

Al2Klimov:
But the first one tests serial adds w/o reset.

oxzi (approved):

PR #48, opened 2024-07-24
logging: Sanitize Journald Field Key Names

oxzi:
As seen over at Icinga Notifications[0], the use of non-alphanumeric
    characters for journaldCore causes fields to be silently discarded.

    This was due to journald's field key validation, which is unfortunately
    not very well documented. Looking at its implementation[1], this library
    code could be adapted to ensure that valid field keys are always used.

    [0]: https://github.com/Icinga/icinga-notifications/pull/254
    [1]: https://github.com/systemd/systemd/blob/11d5e2b5fbf9f6bfa5763fd45b56829ad4f0777f/src/libsystemd/sd-journal/journal-file.c#L1703

### DIFF
--- a/logging/journald_core.go
+++ b/logging/journald_core.go
@@ -5,7 +5,6 @@ import (
 	"github.com/pkg/errors"
 	"github.com/ssgreg/journald"
 	"go.uber.org/zap/zapcore"
-	"strings"
 )
 
 // priorities maps zapcore.Level to journal.Priority.
@@ -25,15 +24,13 @@ func NewJournaldCore(identifier string, enab zapcore.LevelEnabler) zapcore.Core 
 	return &journaldCore{
 		LevelEnabler: enab,
 		identifier:   identifier,
-		identifierU:  strings.ToUpper(identifier),
 	}
 }
 
 type journaldCore struct {
 	zapcore.LevelEnabler
-	context     []zapcore.Field
-	identifier  string
-	identifierU string
+	context    []zapcore.Field
+	identifier string
 }
 
 func (c *journaldCore) Check(ent zapcore.Entry, ce *zapcore.CheckedEntry) *zapcore.CheckedEntry {
@@ -62,6 +59,7 @@ func (c *journaldCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
 	}
 
 	enc := zapcore.NewMapObjectEncoder()
+	// Ensure that all field keys are valid journald field keys. If in doubt, use encodeJournaldFieldKey.
 	c.addFields(enc, fields)
 	c.addFields(enc, c.context)
 	enc.Fields["SYSLOG_IDENTIFIER"] = c.identifier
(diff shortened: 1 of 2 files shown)


oxzi:
Based on @julianbrost's comment <https://github.com/Icinga/icinga-notifications/pull/254#issuecomment-2249722077> at the other PR, I have taken another look. As it turns out, previously I blamed the wrong part, however, I was very self-confident doing so.

After some further investigation, which I also commented over there, <https://github.com/Icinga/icinga-notifications/pull/254#issuecomment-2249859977>, I came to the following conclusion:
> Short update: it's not the `SYSLOG_IDENTIFIER`, but using logger's the name, which is then used to set the `SYSLOG_IDENTIFIER` as prefix for each journald field.
> 
> Unfortunately, the field key is not really specified. For starters, the [Native Journal Protocol](https://systemd.io/JOURNAL_NATIVE_PROTOCOL/) documentation only defines those keys as "environment-like". The [systemd.journal-fields man page](https://www.freedesktop.org/software/systemd/man/latest/systemd.journal-fields.html) shows a multitude of potential keys, but has no specification either. After some digging, I found it in the source, [`…/libsystemd/sd-journal/journal-file.c`](https://github.com/systemd/systemd/blob/11d5e2b5fbf9f6bfa5763fd45b56829ad4f0777f/src/libsystemd/sd-journal/journal-file.c#L1703).
> 
> In a nutshell:
> 
> * Key length MUST be within (0, 64] characters.
> * Key MUST NOT start with `_`, unless it is an protected variable.
> * Key MUST NOT start with a digit.
> * Key MUST only contain `[A-Z0-9_]`.

Thus, i have changed the logic here and introduced a function to ensure that only valid field keys are being set. Please refer to the updated commit message.

Using this commit as the icinga-go-library version in icinga-notification seems to work as intended. Take a look at the following (redacted) `journalctl -o json` output:

```
{
  "PRIORITY": "4",
  "_AUDIT_SESSION": "2",
  "_GID": "100",
  "_UID": "1000",
  "_SYSTEMD_USER_SLICE": "-.slice",
  "_SYSTEMD_OWNER_UID": "1000",
  "ICINGA_NOTIFICATIONS_ERROR": "dial tcp [::1]:5432: connect: connection refused",
  "_COMM": "icinga-notifica",
  "_SYSTEMD_CGROUP": "/user.slice/user-1000.slice/user@1000.service/tmux-spawn-be8f4c8f-da78-4934-af4e-f9de076f568f.scope",
  "_SELINUX_CONTEXT": "kernel",
  "SYSLOG_IDENTIFIER": "icinga-notifications",
  "_SOURCE_REALTIME_TIMESTAMP": "1721903588451251",
  "_CMDLINE": "./icinga-notifications --config config.yml",
  "_RUNTIME_SCOPE": "system",
  "_AUDIT_LOGINUID": "1000",
  "MESSAGE": "database: Can't connect to database. Retrying",
  "_TRANSPORT": "journal",
  "_SYSTEMD_UNIT": "user@1000.service",
  "_PID": "50939",
  "_CAP_EFFECTIVE": "0",
  "_SYSTEMD_SLICE": "user-1000.slice",
  "_EXE": "__/git/github.com/Icinga/icinga-notifications/icinga-notifications"
}
```

logging/journald_core.go
```diff
@@ -64,7 +61,7 @@ func (c *journaldCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
 	enc := zapcore.NewMapObjectEncoder()
 	c.addFields(enc, fields)
 	c.addFields(enc, c.context)
-	enc.Fields["SYSLOG_IDENTIFIER"] = c.identifier
+	enc.Fields[journaldFieldEncode("SYSLOG_IDENTIFIER")] = c.identifier
```
julianbrost, line 64: `SYSLOG_IDENTIFIER` is a field specified in https://www.freedesktop.org/software/systemd/man/latest/systemd.journal-fields.html, so I don't think there's really a point in calling `journaldFieldEncode()` on it. If it would change it, it would set the wrong field afterwards. If anything, I'd put it in a constant and then add `journaldFieldEncode(fieldSyslogIdentifier) == fieldSyslogIdentifier` as a test case.

logging/journald_core_test.go
```diff
@@ -0,0 +1,38 @@
...
+		name   string
+		input  string
+		output string
+	}{
+		{"empty", "", "EMPTY_KEY"},
+		{"lowercase", "foo", "FOO"},
+		{"uppercase", "FOO", "FOO"},
+		{"dash", "foo-bar", "FOO_BAR"},
+		{"non ascii", "snow_☃", "SNOW__"},
+		{"leading numb", "23", "ESC_23"},
```
julianbrost, line 20: There should be enough room here to spell out "number" here.

julianbrost (changes requested):

logging/journald_core.go
oxzi: The idea behind wrapping the field key with a `journaldFieldEncode` call was to signalize that field keys should be sanitized. I am totally aware that this static string is valid, but when someone else will alter this code in the future, they might miss the need to sanitize keys, as the other call to `journaldFieldEncode` happens in the `addFields` method.

oxzi: I have added a comment explaining the necessity to have "clean" field keys. This should be even more direct.

julianbrost (approved):

logging/journald_core.go
```diff
@@ -74,11 +72,55 @@ func (c *journaldCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
...
+	keyParts := []rune(strcase.ScreamingSnake(key))
+	for i, r := range keyParts {
+		if isAlpha(r) || isDigit(r) || r == '_' {
+			continue
+		}
+		keyParts[i] = '_'
+	}
+	key = string(keyParts)
+
+	if !isAlpha(rune(key[0])) {
```
lippserd, line 115: If I'm not mistaken, Go already provides utilities for this:

```suggestion
	key := strcase.ScreamingSnake(key)
	for i, r := range key {
		if !(unicode.IsUpper(r) || unicode.IsDigit(r) || r == '_') {
			continue
		}

		key[i] = '_'
	}

	if !unicode.IsUpper(rune(key[0])) {
```

logging/journald_core.go
```diff
@@ -74,11 +72,55 @@ func (c *journaldCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
...
+// keys a necessity. Unfortunately, there was no specific documentation about the field key syntax available. This
+// function follows the logic enforced in systemd's journal_field_valid function[0].
+//
+// This boils down to:
+// - Key length MUST be within (0, 64] characters.
+// - Key MUST start with [A-Z].
+// - Key characters MUST be [A-Z0-9_].
+//
+//	[0]: https://github.com/systemd/systemd/blob/11d5e2b5fbf9f6bfa5763fd45b56829ad4f0777f/src/libsystemd/sd-journal/journal-file.c#L1703
+func journaldFieldEncode(key string) string {
```
lippserd, line 96: I would rename that function to `encodeJournaldField`.

logging/journald_core.go
julianbrost: journald requires ASCII A-Z, whereas (as the package name already suggests), `unicode.IsUpper()` also returns true for non-ASCII uppercase characters like `Ä`: https://go.dev/play/p/ayqkoWoCL8E

`isUpper` would still be a better name than `isAlpha` though.

oxzi: Setting `key[i] = '_'` in a string is not supported. To do so, it needs to be converted to a `[]rune` or some shenanigans like `key[:i] + "_" + key[i+1:]` are necessary. However, the second option messes up some Unicode characters being bigger than a byte. For example: <https://go.dev/play/p/qo3nfS83ybc>

julianbrost (approved):

logging/journald_core.go
lippserd: > journald requires ASCII A-Z, whereas (as the package name already suggests), `unicode.IsUpper()` also returns true for non-ASCII uppercase characters like `Ä`

My bad, I didn't expect UTF-8 strings to be supported and overlooked the test for it.

lippserd (approved):

PR #50, opened 2024-07-29
redis: Configurable Database

oxzi:
A Redis server might support multiple databases, allowing to use the same Redis for different applications. The database can either be configured for the configuration or later changed via [SELECT][0]. By default, the database 0 will be used.

While both the used go-redis library and even [Icinga 2][1] support this feature, we should expose this here and make this configurable.

As discussed in [Icinga DB issue #676][2], this is not something one might want to encourage all users to use. However, there are definitely setups where this option might come in handy.

[0]: https://redis.io/docs/latest/commands/select/
[1]: https://github.com/Icinga/icinga2/blob/bca1a8447af26313123183efa863f947f9590731/lib/icingadb/icingadb.ti#L23
[2]: https://github.com/Icinga/icingadb/issues/676

### DIFF
--- a/redis/config.go
+++ b/redis/config.go
@@ -45,6 +45,7 @@ type Config struct {
 	Host       string     `yaml:"host"`
 	Port       int        `yaml:"port"`
 	Password   string     `yaml:"password"`
+	Database   int        `yaml:"database" default:"0"`
 	TlsOptions config.TLS `yaml:",inline"`
 	Options    Options    `yaml:"options"`
 }
--- a/redis/client.go
+++ b/redis/client.go
@@ -53,7 +53,7 @@ func NewClientFromConfig(c *Config, logger *logging.Logger) (*Client, error) {
 	options := &redis.Options{
 		Dialer:      dialWithLogging(dialer, logger),
 		Password:    c.Password,
-		DB:          0, // Use default DB,
+		DB:          c.Database,
 		ReadTimeout: c.Options.Timeout,
 		TLSConfig:   tlsConfig,
 	}


redis/config.go
```diff
@@ -45,6 +45,7 @@ type Config struct {
 	Host       string     `yaml:"host"`
 	Port       int        `yaml:"port"`
 	Password   string     `yaml:"password"`
+	Database   int        `yaml:"database" default:"0"`
```
yhabteab, line 48: Please validate the decoded value (`0 =< n < 16`) and reject it if it's invalid, instead of crashing later at runtime.
```bash
2024-08-01T14:41:00.154+0200    FATAL   main    ERR DB index is out of range
can't connect to Redis
main.run
        /Users/yhabteab/Workspace/go/icingadb/cmd/icingadb/main.go:82
main.main
        /Users/yhabteab/Workspace/go/icingadb/cmd/icingadb/main.go:37
runtime.main
        /opt/homebrew/opt/go/libexec/src/runtime/proc.go:271
runtime.goexit
        /opt/homebrew/opt/go/libexec/src/runtime/asm_arm64.s:1222
exit status 1
```

oxzi: As I have written in the [referenced issue](https://github.com/Icinga/icingadb/issues/676#issuecomment-1918641933), "there is no hard limit of sixteen databases. This is just the [default in the configuration](https://raw.githubusercontent.com/redis/redis/7.2/redis.conf)." Thus, I would just let it crash if the user decides to use something outside of the configured Redis scope.

yhabteab (approved):

lippserd (changes requested):
I would also change `GetAddr()` to include the database index, if other than `0`. In Icinga DB, we are using that function to log where we connect to.

oxzi:
> I would also change `GetAddr()` to include the database index, if other than `0`. In Icinga DB, we are using that function to log where we connect to.

I like the idea, but the `GetAddr` method feels like the wrong place for me. During Icinga DB's startup, an implementation of this method is called both for Redis and for the relational database. Both are currently reporting the address of the service. The function name does not match protocol-specific information. Furthermore, no such data is logged for the database.

```
2024-08-01T14:07:41.794Z        INFO    icingadb        Connecting to database at 'mysql.example.com:3306'
2024-08-01T14:07:41.796Z        INFO    icingadb        Connecting to Redis at 'redis.example.com:6380'
```

How about introducing a new method for both `db.DB` and `redis.Client` to report domain specific connection details? Like, for example:

```
2024-08-01T14:07:41.794Z        INFO    icingadb        Connecting to database at "mysql.example.com:3306", type is mysql with user "foo" against database "icingadb"
2024-08-01T14:07:41.796Z        INFO    icingadb        Connecting to Redis at "redis.example.com:6380", with user "bar" against database 23
```

lippserd:
> > I would also change `GetAddr()` to include the database index, if other than `0`. In Icinga DB, we are using that function to log where we connect to.
> 
> I like the idea, but the `GetAddr` method feels like the wrong place for me. During Icinga DB's startup, an implementation of this method is called both for Redis and for the relational database. Both are currently reporting the address of the service. The function name does not match protocol-specific information. Furthermore, no such data is logged for the database.

`GetAddr()` was only introduced for logging, so this method needs to be changed or removed if we do something else. But ...

> **EDIT: I did a thing in #53.**

I will write something there and we can leave this PR as it is.

lippserd (approved):

PR #52, opened 2024-07-30
redis: Configurable Username

oxzi:
The Redis ACL system was introduced with Redis 6.0. It introduced users with precisely granular permissions. This change allows Icinga DB to connect to a Redis with an ACL user.

### DIFF
--- a/redis/config.go
+++ b/redis/config.go
@@ -44,6 +44,7 @@ func (o *Options) Validate() error {
 type Config struct {
 	Host       string     `yaml:"host"`
 	Port       int        `yaml:"port"`
+	Username   string     `yaml:"username"`
 	Password   string     `yaml:"password"`
 	TlsOptions config.TLS `yaml:",inline"`
 	Options    Options    `yaml:"options"`
@@ -53,6 +54,10 @@ type Config struct {
 func (r *Config) Validate() error {
 	if r.Host == "" {
 		return errors.New("Redis host missing")
+	}
+
+	if r.Username != "" && r.Password == "" {
+		return errors.New("Redis password must be set, if username is provided")
 	}
 
 	return r.Options.Validate()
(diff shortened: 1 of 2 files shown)


redis/config.go
```diff
@@ -55,5 +56,9 @@ func (r *Config) Validate() error {
 		return errors.New("Redis host missing")
 	}
 
+	if r.Username != "" && r.Password == "" {
+		return errors.New("A Redis username requires a password as well")
```
yhabteab, line 60: GO errors always begin in lower case, so I would rephrase this to `'password' must be set, if username is provided` and get rid of the `A ...` part.

yhabteab (approved):

redis/config.go
oxzi: Good catch, thanks.

lippserd: I think we should always put important information at the beginning of the sentence. The user has to perform an action, so I opt for Yonas' suggestion.

redis/config.go
```diff
@@ -55,5 +56,9 @@ func (r *Config) Validate() error {
 		return errors.New("Redis host missing")
 	}
 
+	if r.Username != "" && r.Password == "" {
+		return errors.New("password must be set, if username is provided")
```
yhabteab, line 60: Sorry, but after testing this out, the validation error on its own doesn't seem clear enough as to which password it refers to, so I would prefix it with a *lowered* `redis`.

oxzi: With _Redis_ being a proper name, I would not put it to lowercase. Furthermore, it is written in an uppercase error just three lines above. I am going to prefix it now with "Redis", unless someone else wants to philosophize more about an error message.

yhabteab (approved):

lippserd (approved):

PR #53, opened 2024-08-02
db.DB/redis.Client: Details in Address for GetAddr

oxzi:
The idea of a more informative connection information string was brought up in [#50](https://github.com/Icinga/icinga-go-library/pull/50#pullrequestreview-2213720430). Currently, at least Icinga DB and Icinga Notifications are logging their connection target after startup. This information, however, is quite vague and was now extended to contain, e.g., if TLS is used, the username, the 

```
INFO    icingadb        Connecting to database at 'mysql://icingadb@localhost:3306/icingadb'
INFO    icingadb        Connecting to Redis at 'redis://localhost:6380'
```

Furthermore, the zapcore.ObjectMarshaler interface was implemented, allowing both *db.DB and *redis.Client to be used as a zap logging field, showing the address.

### DIFF
--- a/database/db.go
+++ b/database/db.go
@@ -18,6 +18,7 @@ import (
 	"github.com/lib/pq"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
 	"net"
@@ -108,6 +109,7 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
 		if utils.IsUnixAddr(c.Host) {
 			config.Net = "unix"
 			config.Addr = c.Host
+			addr = "(" + config.Addr + ")"
 		} else {
 			config.Net = "tcp"
 			port := c.Port
@@ -115,6 +117,7 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
 				port = 3306
 			}
 			config.Addr = net.JoinHostPort(c.Host, fmt.Sprint(port))
+			addr = config.Addr
 		}
 
 		config.DBName = c.Database
@@ -150,7 +153,6 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
 			return unsafeSetSessionVariableIfExists(ctx, conn, "wsrep_sync_wait", fmt.Sprint(c.Options.WsrepSyncWait))
 		}
 
-		addr = config.Addr
 		db = sqlx.NewDb(sql.OpenDB(NewConnector(connector, logger, connectorCallbacks)), MySQL)
 	case "pgsql":
 		uri := &url.URL{
@@ -208,10 +210,21 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
 			return nil, errors.Wrap(err, "can't open pgsql database")
 		}
 
-		addr = utils.JoinHostPort(c.Host, port)
+		if utils.IsUnixAddr(c.Host) {
+			// https://www.postgresql.org/docs/17/runtime-config-connection.html#GUC-UNIX-SOCKET-DIRECTORIES
+			addr = fmt.Sprintf("(%s/.s.PGSQL.%d)", strings.TrimRight(c.Host, "/"), port)
+		} else {
+			addr = utils.JoinHostPort(c.Host, port)
+		}
 		db = sqlx.NewDb(sql.OpenDB(NewConnector(connector, logger, connectorCallbacks)), PostgreSQL)
 	default:
 		return nil, unknownDbType(c.Type)
+	}
+
+	if c.TlsOptions.Enable {
+		addr = fmt.Sprintf("%s+tls://%s@%s/%s", c.Type, c.User, addr, c.Database)
+	} else {
+		addr = fmt.Sprintf("%s://%s@%s/%s", c.Type, c.User, addr, c.Database)
 	}
 
 	db.SetMaxIdleConns(c.Options.MaxConnections / 3)
@@ -229,9 +242,20 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
 	}, nil
 }
 
-// GetAddr returns the database host:port or Unix socket address.
+// GetAddr returns a URI-like database connection string.
+//
+// It has the following syntax:
+//
+//	type[+tls]://user@host[:port]/database
 func (db *DB) GetAddr() string {
 	return db.addr
+}
+
+// MarshalLogObject implements [zapcore.ObjectMarshaler], adding the database address [DB.GetAddr] to each log message.
+func (db *DB) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddString("database_address", db.GetAddr())
+
+	return nil
 }
 
 // BuildColumns returns all columns of the given struct.
--- a/database/db_test.go
+++ b/database/db_test.go
@@ -1,0 +1,124 @@
+package database
+
+import (
+	"github.com/icinga/icinga-go-library/config"
+	"github.com/icinga/icinga-go-library/logging"
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
+	"testing"
+)
+
+func TestNewDbFromConfig_GetAddr(t *testing.T) {
+	tests := []struct {
+		name string
+		conf *Config
+		addr string
+	}{
+		{
+			name: "mysql-simple",
+			conf: &Config{
+				Type:     "mysql",
+				Host:     "example.com",
+				Database: "db",
+				User:     "user",
+			},
+			addr: "mysql://user@example.com:3306/db",
+		},
+		{
+			name: "mysql-custom-port",
+			conf: &Config{
+				Type:     "mysql",
+				Host:     "example.com",
+				Port:     1234,
+				Database: "db",
+				User:     "user",
+			},
+			addr: "mysql://user@example.com:1234/db",
+		},
+		{
+			name: "mysql-tls",
+			conf: &Config{
+				Type:       "mysql",
+				Host:       "example.com",
+				Database:   "db",
+				User:       "user",
+				TlsOptions: config.TLS{Enable: true},
+			},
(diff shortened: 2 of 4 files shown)


lippserd (changes requested):
I like the idea of informative connection logs, but I would combine that into a new function `MustConnect(context.Context, *logging.Logger)` for both `database.DB` and `Redis.Client`. Ultimately, we do the same thing in many projects, i.e. we log where we connect and fail if it doesn't work. `GetAddr()` can be removed in this PR. And instead of "... against database" I would write "... to database".

We could also consider technical strings, e.g. `[type[+tls]://]usersname@addr/db`: `mysql+tls://kubernetes@locahost:3306/kubernetes`, `user@redis/5`.

oxzi:
> We could also consider technical strings, e.g. `[type[+tls]://]usersname@addr/db`: `mysql+tls://kubernetes@locahost:3306/kubernetes`, `user@redis/5`.

I actually liked this approach even more. Thus, I have re-introduced the `GetAddr` method, but now containing all information to produce strings like those. Furthermore, the `zapcore.ObjectMarshaler` was implemented, allowing the types to be used as zap fields.

Building Icinga DB with this branch results a way more informative output:

```
INFO    icingadb        Connecting to database at 'mysql://icingadb@localhost:3306/icingadb'
INFO    icingadb        Connecting to Redis at 'redis://localhost:6380'
```

database/db_test.go
```diff
@@ -0,0 +1,103 @@
...
+		},
+		{
+			name: "mysql-unix-domain-socket",
+			conf: &Config{
+				Type:     "mysql",
+				Host:     "/var/empty/mysql.sock",
+				Database: "db",
+				User:     "user",
+			},
+			addr: "mysql://user@/var/empty/mysql.sock/db",
```
julianbrost, line 57: This demonstrates one limitation of the approach: `/` can appear both in the URL syntax and in a path. The somewhat proper but ugly solution to this would be to just use `net/url` where this would [result](https://go.dev/play/p/6zYV5SLINe7) in `mysql://user@%2Fvar%2Fempty%2Fmysql.sock/db`.

A nicer looking alternative might be to use `%q` instead of `%s` for unix domain sockets, but that raises the question why make it look like an URL if it actually isn't one.

And a somewhat strange-looking alternative might be to take some inspiration from `lib/pq`: https://github.com/Icinga/icinga-go-library/blob/54b737a484a47834d2cb30015b28fc608ad8d8aa/database/db.go#L160-L163

oxzi: > but that raises the question why make it look like an URL if it actually isn't one.

This was inspired by the comment above, <https://github.com/Icinga/icinga-go-library/pull/53#pullrequestreview-2215105352>.

I don't know what exactly would be seen as a beautiful solution. Maybe make it more strange by prefixing a filesystem path with `file://`. I just checked against the [MySQL documentation](https://dev.mysql.com/doc/refman/8.4/en/connecting-using-uri-or-key-value-pairs.html) and found other beauties like `root@localhost?socket=%2Ftmp%2Fmysql.sock` or `root@localhost?socket=(/tmp/mysql.sock)`.

database/db.go
```diff
@@ -217,15 +224,21 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
...
 		columnMap:       NewColumnMap(db.Mapper),
-		addr:            addr,
+		addrDescription: addrDescription,
 		logger:          logger,
 		tableSemaphores: make(map[string]*semaphore.Weighted),
 	}, nil
 }
 
-// GetAddr returns the database host:port or Unix socket address.
+// GetAddr returns the Redis connection address in a technical form.
```
lippserd, line 233: `database` not `Redis` but I would extend it anyway:

GetAddr returns the URI-like database connection string with the following syntax `type[+tls]://user@host[:port]/database`.

database/db.go
```diff
@@ -217,15 +224,21 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
...
 func (db *DB) GetAddr() string {
-	return db.addr
+	return db.addrDescription
+}
+
+// MarshalLogObject implements zapcore.ObjectMarshaler, adding the database address to each log message.
```
lippserd, line 238: I would link to `GetAttr` here so that it's clear which additional information is logged.

database/db.go
```diff
@@ -217,15 +224,21 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
...
+func (db *DB) MarshalLogObject(encoder zapcore.ObjectEncoder) error {
+	encoder.AddString("database_address", db.addrDescription)
+	return nil
```
lippserd, line 241: In general, please add a newline before return as it improves readability.

I know that there is no specific convention or rule in Go regarding the use of newlines before return statements. Even the official Go style guide does not explicitly mention anything in this regard. But we do this in PHP as well and the majority of the code in Icinga DB and Icinga Go library already does this, so I would make this a convention for us.

database/db_test.go
lippserd: > > but that raises the question why make it look like an URL if it actually isn't one.
> 
> This was inspired by the comment above, [#53 (review)](https://github.com/Icinga/icinga-go-library/pull/53#pullrequestreview-2215105352).

With my comment I didn't mean to imply that it is absolutely necessary to create URI-like strings, but 
since both MySQL and PostgreSQL use URI-like connection strings, I think it's a valid approach to create a relaxed version for logging. For sockets, I would use `(/path/to/socket.sock)` or just leave it as it is.

lippserd (changes requested):

database/db.go
oxzi: Does this really improves readability for a two line function? As long as it's unique all over the code, I am fine.

database/db_test.go
oxzi: I changed it to encapsulate a Unix domain socket in brackets, as it is also supported via MySQL.

database/db.go
```diff
@@ -208,12 +210,18 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
 			return nil, errors.Wrap(err, "can't open pgsql database")
 		}
 
-		addr = utils.JoinHostPort(c.Host, port)
 		db = sqlx.NewDb(sql.OpenDB(NewConnector(connector, logger, connectorCallbacks)), PostgreSQL)
+		addr = utils.JoinHostPort(c.Host, port)
```
julianbrost, line 214: `c.Host` may also contain the path to a Unix domain socket for PostgreSQL. Have you tested that? In any case, adding a test for that would be nice.

Bonus points for correctly handling PostgreSQL's weirdness regarding Unix domain sockets (you configure `/run/postgresql` for example, but the actual socket is `/run/postgresql/.s.PGSQL.5432` and depends on the configured port number, so the port number is actually relevant in this case).

julianbrost (changes requested):

database/db.go
oxzi: Indeed, Unix domain support was missing for PostgreSQL. I have implemented the same quirky logic as for MySQL and have added a test.

> Bonus points for correctly handling PostgreSQL's weirdness

Sorry, but I don't want to implement anything more weird. The current state reflects the configuration as read from the configuration file and should be enough to show users where Icinga DB wants to connect to.

julianbrost: > The current state reflects the configuration as read from the configuration file and should be enough to show users where Icinga DB wants to connect to.

Not entirely. Take the following config as an example. On might hope that the `port:` line is irrelevant, but in case of PostgreSQL, it isn't:

```yaml
database:
  type: pgsql
  host: /run/postgresql
  port: 3306
```

With just the address from this PR logged, that could give you some nasty debugging. However, the error message from `lib/pq` might actually save you as this includes the full socket path (output from the current Icinga DB `main` branch, that error should stay the same though):

```
WARN	database	Can't connect to database. Retrying	{"error": "dial unix /run/postgresql/.s.PGSQL.3306: connect: no such file or directory"}
```

> Sorry, but I don't want to implement anything more weird.

I mean it could be as simple as `(/run/postgresql):5432`. That wouldn't really add any complexity to the implementation.

oxzi: Fair enough, I have taken a closer look at the [documentaiton](https://www.postgresql.org/docs/17/runtime-config-connection.html#GUC-UNIX-SOCKET-DIRECTORIES) and decided to mimic their logic.

database/db.go
```diff
@@ -229,11 +242,18 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
 	}, nil
 }
 
-// GetAddr returns the database host:port or Unix socket address.
+// GetAddr returns a URI-like database connection string of the following syntax: type[+tls]://user@host[:port]/database
```
lippserd, line 245: We could render the syntax as code:

```suggestion
// GetAddr returns a URI-like database connection string of the following syntax:
//
//	type[+tls]://user@host[:port]/database
```

Screenshot 2024-10-07 at 12 58 10

But I don't care too much 😉. (Same for Redis.)

database/db.go
```diff
@@ -229,11 +242,18 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
...
 func (db *DB) GetAddr() string {
 	return db.addr
 }
 
+// MarshalLogObject implements zapcore.ObjectMarshaler, adding the database address, DB.GetAddr, to each log message.
```
lippserd, line 250: Even if this function is only used indirectly, I would add square brackets to `DB.GettAttr`, i.e. `[DB.GetAddr]` so that it is transformed to a [link](https://tip.golang.org/doc/comment#links) in the documentation. Same for Redis.

redis/client.go
```diff
@@ -79,9 +80,32 @@ func NewClientFromConfig(c *Config, logger *logging.Logger) (*Client, error) {
...
+	}
+	if utils.IsUnixAddr(c.Client.Options().Addr) {
+		description += "(" + c.Client.Options().Addr + ")"
+	} else {
+		description += c.Client.Options().Addr
+	}
+	if db := c.Client.Options().DB; db != 0 {
+		description += fmt.Sprintf("/%d", db)
+	}
+	return description
```
lippserd, line 101: Please add a newline before `return`.

lippserd (changes requested):

lippserd (approved):

julianbrost (approved):

PR #58, opened 2024-08-06
Tests for Introduce config.FromEnv()

oxzi:
For #41, as just requested by @Al2Klimov in <https://github.com/Icinga/icinga-go-library/pull/41#issuecomment-2269101306>.

### DIFF
--- a/logging/config.go
+++ b/logging/config.go
@@ -5,11 +5,38 @@ import (
 	"github.com/pkg/errors"
 	"go.uber.org/zap/zapcore"
 	"os"
+	"strings"
 	"time"
 )
 
 // Options define child loggers with their desired log level.
 type Options map[string]zapcore.Level
+
+// UnmarshalText implements encoding.TextUnmarshaler to allow Options to be parsed by env.
+//
+// This custom TextUnmarshaler is necessary as - for the moment - env does not support map[T]encoding.TextUnmarshaler.
+// After <https://github.com/caarlos0/env/pull/323> got merged and a new env release was drafted, this method can be
+// removed.
+func (o *Options) UnmarshalText(text []byte) error {
+	optionsMap := make(map[string]zapcore.Level)
+
+	for _, entry := range strings.Split(string(text), ",") {
+		key, valueStr, found := strings.Cut(entry, ":")
+		if !found {
+			return fmt.Errorf("entry %q cannot be unmarshalled as an Option entry", entry)
+		}
+
+		valueLvl, err := zapcore.ParseLevel(valueStr)
+		if err != nil {
+			return fmt.Errorf("entry %q cannot be unmarshalled as level, %w", entry, err)
+		}
+
+		optionsMap[key] = valueLvl
+	}
+
+	*o = optionsMap
+	return nil
+}
 
 // Config defines Logger configuration.
 type Config struct {
--- a/redis/config_test.go
+++ b/redis/config_test.go
@@ -1,0 +1,100 @@
+package redis
+
+import (
+	"github.com/creasty/defaults"
+	"github.com/icinga/icinga-go-library/config"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func TestConfig(t *testing.T) {
+	var defaultOptions Options
+	require.NoError(t, defaults.Set(&defaultOptions), "setting default options")
+
+	subtests := []struct {
+		name     string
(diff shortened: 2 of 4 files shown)


config/config_test.go
```diff
@@ -44,42 +43,26 @@ func (prefixValidator) Validate() error {
...
+		{"nonptr", EnvOptions{}, simpleValidator{}, true},
+		{"nilptr", EnvOptions{}, (*simpleValidator)(nil), true},
+		{"defaulterr", EnvOptions{}, new(nonStructValidator), true},
+		{"parseeerr", EnvOptions{Environment: map[string]string{"FOO": "bar"}}, &simpleValidator{}, true},
+		{"invalid", EnvOptions{Environment: map[string]string{"FOO": "23"}}, &simpleValidator{}, true},
+		{"simple", EnvOptions{Environment: map[string]string{"FOO": "42"}}, &simpleValidator{42}, false},
+		{"default", EnvOptions{}, &defaultValidator{42}, false},
+		{"override", EnvOptions{Environment: map[string]string{"FOO": "23"}}, &defaultValidator{23}, false},
+		{"prefix", EnvOptions{Environment: map[string]string{"PREFIX_FOO": "42"}, Prefix: "PREFIX_"}, &simpleValidator{42}, false},
+		{"nested", EnvOptions{Environment: map[string]string{"PREFIX_FOO": "42"}}, &prefixValidator{simpleValidator{42}}, false},
```
Al2Klimov, line 61: While on it, you don't have to rebase. I've hard-reset my branch to this ☝️ first commit.

redis/config_test.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/icinga/icinga-go-library/config"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+func TestConfig(t *testing.T) {
+	var defaultOptions Options
+	if err := defaults.Set(&defaultOptions); err != nil {
+		require.NoError(t, err, "setting default options")
```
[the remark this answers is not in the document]
oxzi: Even when a potential error of `defaults.Set(&defaultOptions)` is not directly part of the tests, it will influence the test output. Most test cases using `Options: defaultOptions` would fail the `require.Equal(…)` check, resulting in somebody debugging the actual test, unaware that a precondition failed. Thus, i would like to fail early with a distinguishable error, easing the debugging for future eyes.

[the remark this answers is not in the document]
Al2Klimov: Sure, the code is legit. But you could `s/err != nil/true/` w/o actually changing anything. Hence, the _conditional_ execution is a no-op.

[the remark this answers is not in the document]
oxzi: Now I am getting your point. Changed.

logging/config.go
```diff
@@ -5,12 +5,35 @@ import (
...
 	"os"
+	"strings"
 	"time"
 )
 
 // Options define child loggers with their desired log level.
 type Options map[string]zapcore.Level
 
+// UnmarshalText implements encoding.TextUnmarshaler to allow Options to be parsed by env.
+func (o *Options) UnmarshalText(text []byte) error {
```
Al2Klimov, line 16: It turns out the lib already supports `[]encoding.TextUnmarshaler`, so why not `map[T]encoding.TextUnmarshaler`?

* https://github.com/caarlos0/env/pull/323

oxzi: Thanks for trying to solve this issue upstream. I have updated the PR to reflect this potentially upcoming change in the method's signature, <https://github.com/Icinga/icinga-go-library/compare/7d7fd247cd9a69fddf46f237a6ae54ba475cb7e5..2ba2d8b19d9a9460f7f58cc2cd8d31561c03edf3>.

However, I am against waiting for this PR to get merged and a new env release to get drafted. If this will be supported in the future, the custom `encoding.TextUnmarshaler` can just be removed on our end.

oxzi: As written over at your env PR, it works fine. Thanks again.

However, I would still say to merge it in the current state, eventually removing the `encoding.TextUnmarshaler` implementation later on. We have no "influence" on the upstream, resulting in blocking this PR, resulting in blocking your PR, resulting in having less tests (your PR would be nice to have in #60) and also resulting in blocking the icinga-go-library release.

Al2Klimov (approved):
* I'd prefer https://github.com/caarlos0/env/pull/323, but it seems it is how it is.

PR #59, opened 2024-08-08
Actually retry every database error, including "pq: ..."

Al2Klimov:
They aren't of type *pq.Error, but they're clearly database errors.

closes Icinga/icingadb#620
("fixes" will be a dependency update.)

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -9,6 +9,7 @@ import (
 	"github.com/pkg/errors"
 	"io"
 	"net"
+	"strings"
 	"syscall"
 	"time"
 )
@@ -194,6 +195,13 @@ func Retryable(err error) bool {
 	var mye *mysql.MySQLError
 	var pqe *pq.Error
 	if errors.As(err, &mye) || errors.As(err, &pqe) {
+		return true
+	}
+
+	// For errors without a five-digit code, github.com/lib/pq uses fmt.Errorf().
+	// This returns an unexported error type prefixed with "pq: "
+	// Until this gets changed upstream <https://github.com/lib/pq/issues/1169>, we can only check the error message.
+	if strings.HasPrefix(err.Error(), "pq: ") {
 		return true
 	}
 


retry/retry.go
```diff
@@ -193,7 +194,7 @@ func Retryable(err error) bool {
 
 	var mye *mysql.MySQLError
 	var pqe *pq.Error
-	if errors.As(err, &mye) || errors.As(err, &pqe) {
+	if errors.As(err, &mye) || errors.As(err, &pqe) || strings.HasPrefix(err.Error(), "pq: ") {
```
oxzi, line 197: Could you please add a bit more information here? Does this case necessary belong to this `if`?

Personally, I am a bit puzzled by this change.

Al2Klimov: > This definitely isn't how best-practice error handling in Go should look like,

Not so loud, please! 🙈 https://github.com/Icinga/icingadb/blob/v1.2.0/pkg/icingaredis/utils.go#L96

julianbrost: That's just returning an error, not error handling. Should we have the need to handle that error specifically, that should be changed to its own type rather than checking for that string, that's all.

oxzi: > No, I didn't found any: https://github.com/lib/pq/issues?q=is%3Aissue+is%3Aopen+errorf

Now there's one: https://github.com/lib/pq/issues/1169

oxzi (approved):
While still being unhappy about how the error needs to be checked, the change looks good to me now.

yhabteab:
Please rebase!

yhabteab (approved):

lippserd (approved):

retry/retry.go
```diff
@@ -197,5 +198,11 @@ func Retryable(err error) bool {
 		return true
 	}
 
+	// For errors without a five-digit code, github.com/lib/pq uses fmt.Errorf().
+	// That returns an unexported error type, so we can only check the error message:
```
lippserd, line 202: Please link the created pq issue and also reference it in the commit description.

lippserd (changes requested):

lippserd (approved):

PR #63, opened 2024-08-14
Introduce `DB#ExecTx()` method

yhabteab:
Introduces a `DB#ExecTx()` method on the `DB` receiver (cherry-picked from #18)

### DIFF
--- a/database/db.go
+++ b/database/db.go
@@ -769,6 +769,31 @@ func (db *DB) Delete(
 	return db.DeleteStreamed(ctx, entityType, idsCh, onSuccess...)
 }
 
+// ExecTx executes the provided function within a database transaction.
+//
+// Starts a new transaction, executes the provided function, and commits the transaction
+// if the function succeeds. If the function returns an error, the transaction is rolled back.
+//
+// Returns an error if starting the transaction, executing the function, or committing the transaction fails.
+func (db *DB) ExecTx(ctx context.Context, fn func(context.Context, *sqlx.Tx) error) error {
+	tx, err := db.BeginTxx(ctx, nil)
+	if err != nil {
+		return errors.Wrap(err, "can't start transaction")
+	}
+	// We don't expect meaningful errors from rolling back the tx other than the sql.ErrTxDone, so just ignore it.
+	defer func() { _ = tx.Rollback() }()
+
+	if err := fn(ctx, tx); err != nil {
+		return errors.WithStack(err)
+	}
+
+	if err := tx.Commit(); err != nil {
+		return errors.Wrap(err, "can't commit transaction")
+	}
+
+	return nil
+}
+
 func (db *DB) GetSemaphoreForTable(table string) *semaphore.Weighted {
 	db.tableSemaphoresMu.Lock()
 	defer db.tableSemaphoresMu.Unlock()


database/db.go
```diff
@@ -763,6 +763,26 @@ func (db *DB) Delete(
...
+		return errors.Wrap(err, "DB.RunInTx: cannot start a database transaction")
+	}
+	// We don't expect meaningful errors from rolling back the tx other than the sql.ErrTxDone, so just ignore it.
+	defer func() { _ = tx.Rollback() }()
+
+	if err := fn(ctx, tx); err != nil {
+		return err
+	}
+
+	return errors.Wrap(tx.Commit(), "DB.RunInTx: cannot commit a database transaction")
```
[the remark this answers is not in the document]
yhabteab: > Above in line 773, you have also performed an `err != nil` check first

I did add a nil check there, because if there wasn't an error, I would want to go on with other things, but in this case `tx.Commit()` is the last statement of that method, so I didn't care too much about it, since `errors.Wrap()` also supports nil errors.

oxzi (dismissed):

database/db.go
```diff
@@ -763,6 +763,29 @@ func (db *DB) Delete(
 	return db.DeleteStreamed(ctx, entityType, idsCh, onSuccess...)
 }
 
+// RunInTx allows running a function in a database transaction without requiring manual transaction handling.
+//
+// A new transaction is started on [DB] which is then passed to fn. After fn returns, the transaction is
+// committed unless an error was returned. If fn returns an error, that error is returned or when failing
+// to start or/and commit the transaction.
+func (db *DB) RunInTx(ctx context.Context, fn func(context.Context, *sqlx.Tx) error) error {
```
lippserd, line 771: I would rename that function to `ExecTx`. Isolation level should also be adjustable. Please also consider that we already have special handling for serializable transactions.

database/db.go
```diff
@@ -763,6 +763,29 @@ func (db *DB) Delete(
...
+	tx, err := db.BeginTxx(ctx, nil)
+	if err != nil {
+		return errors.Wrap(err, "DB.RunInTx: cannot start a database transaction")
```
lippserd, line 774: Omit the prefix. Also, `"can't start transaction"` should be totally sufficient because of our error wrapping.

database/db.go
```diff
@@ -763,6 +763,29 @@ func (db *DB) Delete(
...
+	}
+	// We don't expect meaningful errors from rolling back the tx other than the sql.ErrTxDone, so just ignore it.
+	defer func() { _ = tx.Rollback() }()
+
+	if err := fn(ctx, tx); err != nil {
+		return err
+	}
+	if err := tx.Commit(); err != nil {
```
lippserd, line 782: I would add a newline before.

database/db.go
```diff
@@ -763,6 +763,29 @@ func (db *DB) Delete(
...
+func (db *DB) RunInTx(ctx context.Context, fn func(context.Context, *sqlx.Tx) error) error {
+	tx, err := db.BeginTxx(ctx, nil)
+	if err != nil {
+		return errors.Wrap(err, "DB.RunInTx: cannot start a database transaction")
+	}
+	// We don't expect meaningful errors from rolling back the tx other than the sql.ErrTxDone, so just ignore it.
+	defer func() { _ = tx.Rollback() }()
+
+	if err := fn(ctx, tx); err != nil {
+		return err
```
lippserd, line 780: No wrapping here on purpose?

lippserd (changes requested):

database/db.go
yhabteab: What should I wrap it with? Like it states in the method doc block, if the callback returns an error, then it is returned as-is. If in doubt, that provided callback should returned wrapped errors to clearly indicate what's going wrong.

database/db.go
yhabteab: > Please also consider that we already have special handling for serializable transactions.

Which special handling are you referring to? As far as I am aware, our HA tx utilises different method for MySQL driver to simulate a serializable transaction. However, I don't see how this method should make any indication of that.

lippserd: > However, I don't see how this method should make any indication of that.

Why not? I don't understand your answer to be honest.

database/db.go
lippserd: With a stacktrace.

lippserd: > What should I wrap it with?

With a stacktrace.

database/db.go
yhabteab: This method only executes the provided callback and never gets to see what queries you are executing. So how should I append `LOCK IN SHARE MODE` to every single query that is executed within that callback?

database/db.go
yhabteab: Just wrapping with `errors.WithStack()` without additional messages? I can do this, but I just wanted to avoid another layer that meaninglessly wraps the error, which is not helpful for debugging and always expect the callback to return wrapped errors *with context specific messages*.

database/db.go
yhabteab: Therefore, I would just allow to pass `sql.TxOptions` as an argument to this method, and the caller has to perform the special cases required for that callback.

database/db.go
yhabteab: Fixed!

database/db.go
yhabteab: We'll introduce another version of `ExecTx` in the future that takes care of the tx isolation special handling, but this method won't change as it only invokes the provided callback with a newly initialised tx and has no knowledge of the actual queries executed by `fn`.

yhabteab:
Just fixed the outdated commit message which was referring to the now renamed method `DB#RunInTx`.

database/db.go
lippserd: We had a conversation about this, and I wanted to add `errors.WithStack()` because otherwise there is no clear indication that the (error from the) callback actually went through this function.

database/db.go
```diff
@@ -769,6 +769,30 @@ func (db *DB) Delete(
 	return db.DeleteStreamed(ctx, entityType, idsCh, onSuccess...)
 }
 
+// ExecTx allows running a function in a database transaction without requiring manual transaction handling.
```
lippserd, line 772: Description would benefit from simple english and simple tenses, e.g. "allows running a function" vs. "executes", "is started" vs "starts".

I would write the following:

ExecTx executes a function within a database transaction.

Starts a new transaction, executes the provided function, and commits the transaction if the function succeeds. If the function returns an error, the transaction is rolled back.

Returns an error if starting the transaction, executing the function, or committing the transaction fails.

lippserd (changes requested):

database/db.go
```diff
@@ -769,6 +769,30 @@ func (db *DB) Delete(
...
+//
+// A new transaction is started on [DB] which is then passed to fn. After fn returns, the transaction is
+// committed unless an error was returned. If fn returns an error, that error is returned or when failing
+// to start or/and commit the transaction.
+func (db *DB) ExecTx(ctx context.Context, fn func(context.Context, *sqlx.Tx) error) error {
+	tx, err := db.BeginTxx(ctx, nil)
+	if err != nil {
+		return errors.Wrap(err, "cannot start transaction")
```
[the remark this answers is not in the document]
yhabteab: Quite ironic that we have replaced all such abbreviations in Icinga Notifications :-)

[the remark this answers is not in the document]
lippserd: I didn't know that :). If we want to use more formal error messages, I have no problem with that. However, we should then change all error messages in the lib.

lippserd (approved):

yhabteab:
@oxzi do you want to look over the changes again since you previously approved it?

oxzi (approved):

PR #66, opened 2024-08-16
BatchSliceOfStrings(): panic if count <= 0

Al2Klimov:
Otherwise, 0 would lead to infinite empty slices.
Negative numbers already crash the program.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -30,7 +30,12 @@ func Timed(start time.Time, callback func(elapsed time.Duration)) {
 }
 
 // BatchSliceOfStrings groups the given keys into chunks of size count and streams them into a returned channel.
+// Panics if count is less than or equal to zero.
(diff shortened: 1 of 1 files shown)


oxzi (approved):
As I have tested and documented over at <https://github.com/Icinga/icinga-go-library/pull/32#discussion_r1717147418>, both a `count` of zero as well as a negative `count` only do harm.

yhabteab:
Currently it is only used here:
https://github.com/Icinga/icinga-go-library/blob/15703e68fdfce1724833d8af1b8ca977a6cfde71/redis/client.go#L163

and `redis.Options'HMGetCount` is not allowed to be zero or a negative number, so this should be fine.
https://github.com/Icinga/icinga-go-library/blob/15703e68fdfce1724833d8af1b8ca977a6cfde71/redis/config.go#L24-L26

PR #67, opened 2024-08-19
retry.Retryable(): don't re-try context.DeadlineExceeded

Al2Klimov:
It is considered temporary and timeout, but we don't want to retry on it.

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -136,6 +136,11 @@ func ResetTimeout(t *time.Timer, d time.Duration) {
 // i.e. temporary, timeout, DNS, connection refused and reset, host down and unreachable and
 // network down and unreachable errors. In addition, any database error is considered retryable.
 func Retryable(err error) bool {
+	if errors.Is(err, context.DeadlineExceeded) {
+		// context.DeadlineExceeded matches both temporary and timeout below, but we don't want to retry on it.
+		return false
+	}
+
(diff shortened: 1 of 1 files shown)


retry/retry.go
```diff
@@ -136,6 +136,11 @@ func ResetTimeout(t *time.Timer, d time.Duration) {
 // i.e. temporary, timeout, DNS, connection refused and reset, host down and unreachable and
 // network down and unreachable errors. In addition, any database error is considered retryable.
 func Retryable(err error) bool {
+	if errors.Is(err, context.DeadlineExceeded) {
+		// context.DeadlineExceeded is considered temporary and timeout, but we don't want to retry on it.
```
oxzi, line 140: This sentence looks wrong. Is it temporary and what about the timeout part? Could you please rephrase it.

oxzi (commented):
Context for this PR is provided in <https://github.com/Icinga/icinga-go-library/pull/23#discussion_r1705689655>.

oxzi (approved):

yhabteab:
Are you sure this won't break something? There is already a special handling for ctx and network I/O errors here.
https://github.com/Icinga/icinga-go-library/blob/2eac186ed1b91ff4f217194b363c511ea9e1bb19/retry/retry.go#L72-L77

PR #68, opened 2024-09-03
periodic.WindUpContext: Add function

oxzi:
WindUpContext wraps a context.Context with a wind-up timeout, allowing to be extended multiple times. This extends the possibilities of context.WithTimeout with a timeout to be reset multiple times.

This function was implemented to be used in Icinga DB's HA logic.


oxzi:
Closed as no longer needed :smiling_face_with_tear:

PR #69, opened 2024-09-04
retry.WithBackoff: QuickContextExit, Documentation, Tests

oxzi:
The retry.WithBackoff function only loosely respected the provided context. In particular, the RetryableFunc could survive the context termination and block as long as it desires.

For situations where timing is critical and a deadline bound context is used, this behavior might be undesirable. That's why the new Settings.QuickContextExit option has been introduced, which effectively wraps time-consuming, blocking functions with an additional context check and bails out directly when the context is done.

This change was reflected in the documentation, which was generally extended, trying to explain more about the implementation details.

While working on the function, some minor refactorings were done. Notable is the simplified check if the context has finished.

Finally, tests were written for retry.WithBackoff to verify the expected behavior and ease future changes.

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -3,6 +3,7 @@ package retry
 import (
 	"context"
 	"database/sql/driver"
+	"fmt"
 	"github.com/go-sql-driver/mysql"
 	"github.com/icinga/icinga-go-library/backoff"
 	"github.com/lib/pq"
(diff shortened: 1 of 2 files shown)


retry/retry.go
```diff
@@ -30,61 +30,124 @@ type OnSuccessFunc func(elapsed time.Duration, attempt uint64, lastErr error)
...
+
+	// OnRetryableError, if not nil, is called if a retryable error occurred.
 	OnRetryableError OnRetryableErrorFunc
-	OnSuccess        OnSuccessFunc
+
+	// OnSuccess, if not nil, is called after the function succeeded.
+	OnSuccess OnSuccessFunc
+
+	// QuickContextExit, if set, directly aborts if the context is done and does not wait for functions to finish.
+	QuickContextExit bool
```
julianbrost, line 50: From the doc comments, I don't think it's very clear how these two interact:

> RetryableFunc is always granted its full execution time and is not canceled if it exceeds Timeout.

> QuickContextExit, if set, directly aborts if the context is done and does not wait for functions to finish.

retry/retry.go
```diff
@@ -30,61 +30,124 @@ type OnSuccessFunc func(elapsed time.Duration, attempt uint64, lastErr error)
...
+		err = funcWrapper(func(ctx context.Context) error {
+			err := retryableFunc(ctx)
+			if err == nil {
+				if settings.OnSuccess != nil {
+					settings.OnSuccess(time.Since(start), attempt, prevErr)
+				}
 			}
-
+			return err
+		})(ctx)
```
julianbrost, line 131: Is the purpose of calling `OnSuccess` within the wrapped function so that this callback couldn't delay the return of this function either?

However, this seems to move calling the callbacks to another goroutine, which might be surprising and could introduce race conditions.

retry/retry.go
```diff
@@ -30,61 +30,124 @@ type OnSuccessFunc func(elapsed time.Duration, attempt uint64, lastErr error)
...
+// While the function parameter signature matches RetryableFunc, this is not used here as this function does not only
+// focuses on RetryableFunc, but more generally functions which might get aborted within WithBackoff.
+func contextBoundFunc(f func(context.Context) error) func(context.Context) error {
+	return func(ctx context.Context) error {
+		fResultCh := make(chan error)
+
+		go func() {
+			fResultCh <- f(ctx)
+			close(fResultCh)
+		}()
```
julianbrost, line 64: Have you considered what happens if `f` panics? Looks like the rest of the code relies on the context to finish at some point then.

retry/retry.go
```diff
@@ -30,61 +30,124 @@ type OnSuccessFunc func(elapsed time.Duration, attempt uint64, lastErr error)
...
+		select {
+		case err := <-fResultCh:
+			return err
+
+		case <-ctx.Done():
+			go func() {
+				// Drain channel and wait for it to be closed in the background.
+				for range fResultCh {
+				}
+			}()
```
julianbrost, line 75: I think you'd achieve the same by making the channel buffered with size 1 and leave this job to the garbage collector.

retry/retry.go
```diff
@@ -30,61 +30,124 @@ type OnSuccessFunc func(elapsed time.Duration, attempt uint64, lastErr error)
...
+	}
 }
 
-// WithBackoff retries the passed function if it fails and the error allows it to retry.
-// The specified backoff policy is used to determine how long to sleep between attempts.
+// WithBackoff retries the retryableFunc in case of failure.
+//
+// The passed ctx loosely restricts retries and prohibits retries if the context is done. It is also being passed to the
+// retryableFunc, which should honor the context as well. However, unless Settings.QuickContextExit is set, the
+// RetryableFunc can survive after the context has already finished.
```
julianbrost, line 86: Doesn't it also survive if `QuickContextExit` is set, just with the difference that it's running in a different goroutine in the background?

retry/retry.go
oxzi: Yes. Theoretically, both callback functions may block as well, while, unfortunately, they do not even receive a context to be bound to.

I am a bit unhappy with this architecture, but at least it does not break the API and needs a check for every `retry.WithBackoff` call. Another option would be to exclude the callbacks from the context-wrapper and stating this in the documentation. In most cases, the callbacks are only used for logging which should not block.

retry/retry.go
oxzi: It relies that either the function or the context finishes, yes. Otherwise, the `WithBackoff` function will not return as well.

I have added a recovery check, good point.

yhabteab:
Since https://github.com/Icinga/icingadb/pull/800 doesn't make use of this, is it still necessary? If we still intend to include it, it can be shipped in later releases, but it's not required for the `0.4.0` release. Thus, please close or drop it from the milestone based on your decision.

oxzi:
Closed as no longer needed :smiling_face_with_tear:

PR #72, opened 2024-09-19
config/: enforce type constraints via generics, not reflection

Al2Klimov:


config/config.go
```diff
@@ -8,21 +8,25 @@ import (
...
+}
+
 // FromYAMLFile parses the given YAML file and stores the result
 // in the value pointed to by v. If v is nil or not a pointer,
 // FromYAMLFile returns an [ErrInvalidArgument] error.
-func FromYAMLFile(name string, v Validator) error {
-	rv := reflect.ValueOf(v)
-	if rv.Kind() != reflect.Pointer || rv.IsNil() {
-		return errors.Wrapf(ErrInvalidArgument, "non-nil pointer expected, got %T", v)
+func FromYAMLFile[T any, V validatorPtrTo[T]](name string, v V) error {
```
Al2Klimov, line 27:
```go
	FromYAMLFile("", new(int))
	FromYAMLFile("", redis.Config{})
	FromYAMLFile("", &redis.Config{})
```

Only the latter compiles. 👍

config/config.go
```diff
@@ -59,10 +63,9 @@ func FromYAMLFile(name string, v Validator) error {
 // ParseFlags prints the help message to [os.Stdout] and exits.
 // Note that errors are not printed automatically,
 // so error handling is the sole responsibility of the caller.
-func ParseFlags(v any) error {
-	rv := reflect.ValueOf(v)
-	if rv.Kind() != reflect.Pointer || rv.IsNil() {
-		return errors.Wrapf(ErrInvalidArgument, "non-nil pointer expected, got %T", v)
+func ParseFlags[T any](v *T) error {
```
Al2Klimov, line 66:
```go
	ParseFlags(int(1))
	ParseFlags(new(int))
```

Only the latter compiles. 👍

config/config.go
```diff
@@ -8,21 +8,25 @@ import (
...
 // FromYAMLFile parses the given YAML file and stores the result
 // in the value pointed to by v. If v is nil or not a pointer,
 // FromYAMLFile returns an [ErrInvalidArgument] error.
-func FromYAMLFile(name string, v Validator) error {
-	rv := reflect.ValueOf(v)
-	if rv.Kind() != reflect.Pointer || rv.IsNil() {
-		return errors.Wrapf(ErrInvalidArgument, "non-nil pointer expected, got %T", v)
+func FromYAMLFile[T any, V validatorPtrTo[T]](name string, v V) error {
+	if v == nil {
+		return errors.Wrap(ErrInvalidArgument, "got nil pointer")
```
Al2Klimov, line 29: While on it, we could also remove those right away.

config/config.go
```diff
@@ -59,10 +63,9 @@ func FromYAMLFile(name string, v Validator) error {
 // ParseFlags prints the help message to [os.Stdout] and exits.
 // Note that errors are not printed automatically,
 // so error handling is the sole responsibility of the caller.
-func ParseFlags(v any) error {
-	rv := reflect.ValueOf(v)
-	if rv.Kind() != reflect.Pointer || rv.IsNil() {
-		return errors.Wrapf(ErrInvalidArgument, "non-nil pointer expected, got %T", v)
+func ParseFlags[T any](v *T) error {
+	if v == nil {
+		return errors.Wrap(ErrInvalidArgument, "got nil pointer")
```
Al2Klimov, line 68: I mean, which of our functions which take pointers checks them like this?

config/config.go
lippserd: Why should we allow passing nil (pointers) and why is it an argument that this is not checked by other functions? If we omit this check, `defaults.Set()` in `FromYAMLFile()` will panic and `ParseFlags()` will return different errors depending on whether flags were specified on the command line or not (all if nil was specified). So, why?

config/config.go
lippserd: https://github.com/Icinga/icinga-go-library/pull/72/files#r1768083795

config/contracts.go
```diff
@@ -3,3 +3,9 @@ package config
 type Validator interface {
 	Validate() error
 }
+
+// validatorPtrTo combines the [Validator] interface with a pointer constraint.
+type validatorPtrTo[T any] interface {
```
lippserd, line 8: I would name it `validatorPtr`.

lippserd (changes requested):
I already implemented something similar when I did the library splitting, but without passing a parameter. In the end I removed the whole thing because it is more straightforward to actually pass a parameter:

https://github.com/Icinga/icingadb/pull/747#discussion_r1584915453

In general, I like the changes here!

config/config.go
Al2Klimov: It's not only our functions, but also the Go stuff doesn't nil check pointer params, e.g. all methods of atomic.Int64, because those are programming errors. IMAO [Pete explained](https://softwareengineering.stackexchange.com/a/298486) it very well.

lippserd: > e.g. all methods of atomic.Int64

So you are comparing low-level pointer receiver functions with our high level library functions?

> because those are programming errors

True. But why is it a problem to just indicate that with a clear error message in a library? Besides, `ParseFlags()` **does not** panic. It yields random errors on first glance.

>IMAO [Pete explained](https://softwareengineering.stackexchange.com/a/298486) it very well.

Why can't you just summarize it? Why should I read this wall of text? Besides, I would prefer your own arguments to those of a random person who has  been "using [Go] professionally for about 6 months". 🤣

> It's not only our functions, but also the Go stuff doesn't nil check pointer params

Similar functions like `json.Unmarshal()`, `errors.As()`, etc. are a perfect example of how it's done.

lippserd (approved):

PR #73, opened 2024-09-24
database: Document context limitations of Tx.Commit()

oxzi:
A key finding of Icinga/icingadb#800 was that committing a transaction does not necessarily have to respect the context of the transaction, depending on the database driver. As @lippserd suggested there, I have added notes to the documentation of all relevant database functions.

### DIFF
--- a/database/db.go
+++ b/database/db.go
@@ -540,8 +540,11 @@ func (db *DB) NamedBulkExec(
 // Takes in up to the number of entities specified in count from the arg stream and
 // executes a new transaction that runs a new query for each entity in this set of arguments,
 // until the arg stream has been processed.
+//
 // The transactions are executed in a separate goroutine with a weighting of 1
 // and can be executed concurrently to the extent allowed by the semaphore passed in sem.
+//
+// Note that committing the transaction may not honor the context provided, as described further in [DB.ExecTx].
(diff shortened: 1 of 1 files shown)


database/db.go
```diff
@@ -540,8 +540,11 @@ func (db *DB) NamedBulkExec(
 // Takes in up to the number of entities specified in count from the arg stream and
 // executes a new transaction that runs a new query for each entity in this set of arguments,
 // until the arg stream has been processed.
+//
 // The transactions are executed in a separate goroutine with a weighting of 1
 // and can be executed concurrently to the extent allowed by the semaphore passed in sem.
+//
+// Note that committing the transaction may not honor the context provided, as described further in DB.ExecTx.
```
lippserd, line 547: Please add square brackets to `DB.ExecTx`, i.e. `[DB.ExecTx]` so that it is transformed to a [link](https://tip.golang.org/doc/comment#links) in the documentation.

lippserd (changes requested):

database/db.go
oxzi: Always these new-fashioned things, <https://github.com/golang/proposal/blob/master/design/51082-godocfmt.md#new-syntax-for-url-links>.

PR #79, opened 2024-10-21
Revert "Merge pull request #72 from Icinga/reflectPtr"

lippserd:
This reverts commit f25f8a03e301395f9bf3873783970f8e502479a7, reversing changes made to a2fd4d4fc03652d0871e4800eeb2f2b005d6f08b.

While the changes relax the necessary checks for passing a valid pointer, they complicate the tests and still allow non-construct pointer types that implement the validator interface to pass. This negates the benefits, making it simpler to validate the pointer argument in a function as previously done.

refs #72

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -8,6 +8,7 @@ import (
 	"github.com/jessevdk/go-flags"
 	"github.com/pkg/errors"
 	"os"
+	"reflect"
 )
 
 // ErrInvalidArgument is the error returned by [ParseFlags] or [FromYAMLFile] if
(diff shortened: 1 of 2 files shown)


lippserd:
I think `validatorPtr` can be changed to the following to only allow struct pointer:

```golang
// validatorPtr combines the [Validator] interface with a pointer constraint.
type validatorPtr[T ~struct{}] interface {
	Validator
	*T
}
```

Nevertheless, the tests would have to be adapted, and the non-nil checks can't be removed anyway.

yhabteab (approved):

PR #81, opened 2024-10-23
Also validate that arguments to `config` functions are struct pointers

lippserd:
Previously, the libraries used, i.e. `jessevdk/go-flags` and `creasty/defaults`, returned an error if the specified argument was not a struct pointer. We now do this in our validation.

The PR also introduces some minor code quality changes.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -14,16 +14,15 @@ import (
 
 // ErrInvalidArgument is the error returned by [ParseFlags] or [FromYAMLFile] if
 // its parsing result cannot be stored in the value pointed to by the designated passed argument which
-// must be a non-nil pointer.
+// must be a non-nil struct pointer.
 var ErrInvalidArgument = stderrors.New("invalid argument")
 
 // FromYAMLFile parses the given YAML file and stores the result
-// in the value pointed to by v. If v is nil or not a pointer,
+// in the value pointed to by v. If v is nil or not a struct pointer,
 // FromYAMLFile returns an [ErrInvalidArgument] error.
 func FromYAMLFile(name string, v Validator) error {
-	rv := reflect.ValueOf(v)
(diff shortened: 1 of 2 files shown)


oxzi (approved):

config/config.go
```diff
@@ -14,19 +14,18 @@ import (
...
 func FromYAMLFile(name string, v Validator) error {
-	rv := reflect.ValueOf(v)
-	if rv.Kind() != reflect.Pointer || rv.IsNil() {
-		return errors.Wrapf(ErrInvalidArgument, "non-nil pointer expected, got %T", v)
+	if err := validateNonNilStructPointer(v); err != nil {
+		return errors.WithStack(err)
 	}
 
-	// #nosec G304 -- Potential file inclusion via variable - Its purpose is to load any file name that is passed to it, so doesn't need to validate anything.
+	// #nosec G304 -- Accept user-controlled input for config file.
```
yhabteab, line 28: I don't understand the reason for this change. Before, it explained what that magic code stands for without having to google it and why it's necessary to suppress it, but now I can't see how that description relates to the suppressed Gosec code.

lippserd: Format is // nosec - justification. I don’t want to add a description to each and every code. In my opinion the justification is clear enough, i.e. accepting user controlled input. I could add „without validation“ though.

yhabteab: I'm pretty sure I got that format from some gosec guidelines, but I can't find the exact source right now. However, if this is the recommended way/best practice, then it's fine and I don't want to block the PR just because of this.

lippserd: https://github.com/securego/gosec?tab=readme-ov-file#annotating-code

yhabteab: I don't know where I got this from then :)

PR #105, opened 2024-12-09
Add support for sqlite

jrauh01:


database/db.go
```diff
@@ -21,6 +21,7 @@ import (
 	"go.uber.org/zap/zapcore"
 	"golang.org/x/sync/errgroup"
 	"golang.org/x/sync/semaphore"
+	_ "modernc.org/sqlite"
```
oxzi, line 24: While in theory having a CGo-free SQLite database driver sounds tempting, this library introduces a hard limitation on targeted platforms. Please see the [README section](https://pkg.go.dev/modernc.org/sqlite#hdr-Supported_platforms_and_architectures) or this [random issue](https://gitlab.com/cznic/sqlite/-/issues/88). Since this code is introduced in our `database` package, I am unsure how this applies to downstream consumers such as Icinga DB.

So even if CGo is a nuisance, I would recommend using the good old [`github.com/mattn/go-sqlite3` library](https://github.com/mattn/go-sqlite3), as we already do in Icinga DB for the `icingadb-migrate` utility.

oxzi (commented):
First of all, thanks for this PR. While skimming older open PRs, I stumbled upon this one and its `modernc.org/sqlite` dependency.

Besides the one comment about the library, could you please describe the motivation for this change and its necessity? Where is it needed and why?

database/db.go
oxzi: No big deal. Honestly, it's not that obvious. In the past, I would have liked to use this library as well.

database/db.go
```diff
@@ -217,14 +218,34 @@ func NewDbFromConfig(c *Config, logger *logging.Logger, connectorCallbacks Retry
...
+			name = c.Database
+			mode = ""
+		)
+
+		if strings.Contains(c.Database, ":memory:") && c.Database != ":memory:" {
+			name = strings.Split(c.Database, ":memory:")[1]
+			mode = "mode=memory&"
+		}
+
+		addr = fmt.Sprintf("file:%s?%scache=shared", name, mode)
```
oxzi, line 232: As I was just fiddling with some other SQLite application, I realized that by default foreign key constraints are not enforced unless `_foreign_keys` is set accordingly. After fixing this on my side, I remembered this PR and would suggest adding this here as well.

jrauh01:
This PR was thought for a new implementation of the Icinga Kubernetes daemon. This is not planned to be implemented in the near future, so I will close this PR.

PR #113, opened 2025-03-05
config.TLS: Inline PEM-encoded Certs, Keys and CAs

oxzi:
Introduced support for raw encoded PEM data instead of filenames for TLS.Cert, TLS.Key and TLS.Ca. This should make it easier to get rid of the additional entrypoint program for docker-icingadb and allow direct use of Icinga DB with its environment variable support.

In addition to the implementation, new tests were written with the goal of complete coverage. A consumer of this type, database.Config, got another test case using PEMs both as environment variables and as YAML


config/tls.go
```diff
@@ -57,6 +61,9 @@ func (t *TLS) MakeConfig(serverName string) (*tls.Config, error) {
 		return nil, nil
 	}
 
+	isRawPem := func(s string) bool { return strings.HasPrefix(strings.TrimSpace(s), "-----BEGIN") }
```
julianbrost, line 64: Maybe just test if it can be parsed as PEM instead?

PEM files can contain additional data outside these blocks can for example be used to add a comment above each certificate in a file containing multiple certificates.

Should be as simple as that (using [`encoding/pem.Decode()`](https://pkg.go.dev/encoding/pem#Decode)):

```go
block, _ := pem.Decode(s)
return block != nil
```

oxzi:
@lippserd: Valid point. I have moved the decision logic to a helper function like you suggested. Now the body of the actual `MakeConfig` method looks cleaner. While working on it, I have flattened the `if` and removed the nested `if`, making it easier to read IMO.

config/tls.go
```diff
@@ -58,31 +75,37 @@ func (t *TLS) MakeConfig(serverName string) (*tls.Config, error) {
 	}
 
 	tlsConfig := &tls.Config{MinVersion: tls.VersionTLS12}
-	if t.Cert == "" {
-		if t.Key != "" {
-			return nil, errors.New("private key given, but client certificate missing")
-		}
-	} else if t.Key == "" {
+	if t.Cert == "" && t.Key != "" {
```
lippserd, line 78: I would omit the else statements for the early exit conditions:

```golang
if (key-only) {
    return
}

if (cert-only) {
    return
}

if (cert-and-key) {
    ...
}
```

lippserd (changes requested):
I only have one more small thing, otherwise the changes look very good.

config/tls.go
oxzi: Please take another look.

lippserd (approved):

PR #124, opened 2025-04-15
database,redis: More verbosity in OnRetryableError

oxzi:
database,redis: More verbosity in OnRetryableError

    The OnRetryableError functions came with conditions to only log the
    error if it had changed since the last iteration. While this keeps the
    log file clean, it hides reoccurring errors and makes it harder to
    identify specific errors.

    The changed version always logs the reoccurring error and also includes
    the duration since the start and the current attempt.

### DIFF
--- a/database/db.go
+++ b/database/db.go
@@ -879,10 +879,11 @@ func (db *DB) HasTable(ctx context.Context, table string) (bool, error) {
 func (db *DB) GetDefaultRetrySettings() retry.Settings {
 	return retry.Settings{
 		Timeout: retry.DefaultTimeout,
-		OnRetryableError: func(_ time.Duration, _ uint64, err, lastErr error) {
-			if lastErr == nil || err.Error() != lastErr.Error() {
-				db.logger.Warnw("Can't execute query. Retrying", zap.Error(err))
-			}
+		OnRetryableError: func(elapsed time.Duration, attempt uint64, err, lastErr error) {
+			db.logger.Warnw("Can't execute query. Retrying",
+				zap.Error(err),
+				zap.Duration("after", elapsed),
+				zap.Uint64("attempt", attempt))
 		},
 		OnSuccess: func(elapsed time.Duration, attempt uint64, lastErr error) {
 			if attempt > 1 {
--- a/database/driver.go
+++ b/database/driver.go
@@ -67,9 +67,10 @@ func (c RetryConnector) Connect(ctx context.Context) (driver.Conn, error) {
 					c.callbacks.OnRetryableError(elapsed, attempt, err, lastErr)
 				}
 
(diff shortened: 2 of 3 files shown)


oxzi:
Removing "retry deadline exceeded" from the error message also results in more useful error messages for users. Take <https://github.com/Icinga/icingadb/issues/900> for an example.

lippserd (changes requested):
> retry.WithBackoff: Reduce error wrapping
> ...
> But this additional layer does not really carry any useful information.

This helps to easily distinguish non-retryable from "timeout" errors.

> Removing "retry deadline exceeded" from the error message also results in more useful error messages for users. Take [Icinga/icingadb#900](https://github.com/Icinga/icingadb/issues/900) for an example.

That's a different "retry deadline exceeded" error:

https://github.com/Icinga/icingadb/blob/main/pkg/icingadb/ha.go#L166

That doesn't mean that I don't want to change anything but I'm not a fan of just removing the wrapping. The `ha:166` error could get more context, a more understandable message or whatever. And if wrapping is such a problem, I would change the renderer, output the error messages first and then the stack trace(s).

> database/redis: More verbosity in OnRetryableError
> ...
> The changed version always logs the reoccurring error and also includes
the duration since the start and the current attempt.

If I am not mistaken, this leads to a large number of error messages in the first few seconds, as there is not much time between retries. Can this be optimized, for example, to only log ~~every~~ after at least 10 seconds if the error message has not changed?

oxzi:
> > retry.WithBackoff: Reduce error wrapping
> > ...
> > But this additional layer does not really carry any useful information.
> 
> This helps to easily distinguish non-retryable from "timeout" errors.

Fair enough. I just dislike having another entry in the stack trace carrying the position of this wrapping. I am going to remove this.

> > Removing "retry deadline exceeded" from the error message also results in more useful error messages for users. Take [Icinga/icingadb#900](https://github.com/Icinga/icingadb/issues/900) for an example.
> 
> That's a different "retry deadline exceeded" error:
> 
> https://github.com/Icinga/icingadb/blob/main/pkg/icingadb/ha.go#L166

You are totally right. I have successfully confused myself.

> > database/redis: More verbosity in OnRetryableError
> > ...
> > The changed version always logs the reoccurring error and also includes
> > the duration since the start and the current attempt.
> 
> If I am not mistaken, this leads to a large number of error messages in the first few seconds, as there is not much time between retries. Can this be optimized, for example, to only log ~every~ after at least 10 seconds if the error message has not changed?

Yes, this results in more WARNING log entries but this was my intention.

IMO, if something is broken, it should be shown and not hidden. The current behavior only shows an error once, hiding it for up to five minutes before crashing. The suggested behavior would always show the error, resulting in a few entries recurring entries at first - but only if the error is not recovered directly.

PR #126, opened 2025-04-22
logging: Include error field in journald message, Fix key encoding of all fields for journaldCore

oxzi:
### logging: Include error field in journald message

Our journald zapcore.Core logs every zap field as a journald field. This has the disadvantage that these fields do not shown up in the default journalctl output, effectively hiding errors. While this is documented, it is not the expected behavior for most users, resulting in incomplete bug reports.

This change introduces a new set of zap fields to be also included in the journald message. Currently, only the "error" field gets this special treatment, allowing errors to be displayed.

An exemplary Icinga DB log message would change as follows while still containing the ICINGADB_ERROR field.

> database: Can't connect to database. Retrying
> database: Can't connect to database. Retrying	error="dial tcp [::1]:5432: connect: connection refused"

As a drive-by change, the package-global variable "priorities" was renamed to "journaldPriorities" to make it more obvious.

Fixes Icinga/icingadb#793.

---

### logging.journaldCore: Fix key encoding of all fields

The encodeJournaldFieldKey function sanitizes the keys for the journald fields. However, calling the `AddTo` method on some zap fields results in multiple output fields. The keys of these additional fields are not sanitized and may not comply with journald's requirements.

This change applies name sanitation after all fields were added, thereby including also these surplus fields. As a result, errors are now also available as "ERROR_VERBOSE", including a potential stack trace.

### DIFF
--- a/logging/journald_core.go
+++ b/logging/journald_core.go
@@ -1,14 +1,17 @@
 package logging
 
 import (
+	"fmt"
 	"github.com/icinga/icinga-go-library/strcase"
+	"github.com/icinga/icinga-go-library/utils"
 	"github.com/pkg/errors"
 	"github.com/ssgreg/journald"
 	"go.uber.org/zap/zapcore"
+	"strings"
 )
 
-// priorities maps zapcore.Level to journal.Priority.
-var priorities = map[zapcore.Level]journald.Priority{
+// journaldPriorities maps zapcore.Level to journal.Priority.
+var journaldPriorities = map[zapcore.Level]journald.Priority{
 	zapcore.DebugLevel:  journald.PriorityDebug,
 	zapcore.InfoLevel:   journald.PriorityInfo,
 	zapcore.WarnLevel:   journald.PriorityWarning,
@@ -16,6 +19,11 @@ var priorities = map[zapcore.Level]journald.Priority{
 	zapcore.FatalLevel:  journald.PriorityCrit,
 	zapcore.PanicLevel:  journald.PriorityCrit,
 	zapcore.DPanicLevel: journald.PriorityCrit,
+}
+
+// journaldVisibleFields is a set (map to struct{}) of field keys being logged within the message for journald.
+var journaldVisibleFields = map[string]struct{}{
+	"error": {},
 }
 
 // NewJournaldCore returns a zapcore.Core that sends log entries to systemd-journald and
@@ -53,7 +61,7 @@ func (c *journaldCore) With(fields []zapcore.Field) zapcore.Core {
 }
 
 func (c *journaldCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
-	pri, ok := priorities[ent.Level]
+	pri, ok := journaldPriorities[ent.Level]
 	if !ok {
 		return errors.Errorf("unknown log level %q", ent.Level)
 	}
@@ -64,7 +72,7 @@ func (c *journaldCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
 	c.addFields(enc, c.context)
 	enc.Fields["SYSLOG_IDENTIFIER"] = c.identifier
 
-	message := ent.Message
+	message := ent.Message + visibleFieldsMsg(journaldVisibleFields, append(fields, c.context...))
 	if ent.LoggerName != c.identifier {
 		message = ent.LoggerName + ": " + message
 	}
@@ -123,4 +131,57 @@ func encodeJournaldFieldKey(key string) string {
 	}
 
 	return key
+}
+
+// visibleFieldsMsg creates a string to be appended to the log message including fields to be explicitly printed.
+//
+// When logging against journald, the zapcore.Fields are used as journald fields, resulting in not being shown in the
+// default journalctl output (short). While this is documented in our docs, missing error messages are usually confusing
+// for end users.
+//
+// This method takes an allow list (set, map of keys to empty struct) of key to be displayed - there is the global
+// variable journaldVisibleFields; parameter for testing - and a slice of zapcore.Fields, creating an output string of
+// the allowed fields prefixed by a whitespace separator. If there are no fields to be logged, the returned string is
+// empty. So the function output can be appended to the output message without further checks.
+func visibleFieldsMsg(visibleFieldKeys map[string]struct{}, fields []zapcore.Field) string {
+	if visibleFieldKeys == nil || fields == nil {
+		return ""
+	}
+
+	enc := zapcore.NewMapObjectEncoder()
+
+	for _, field := range fields {
+		if _, shouldLog := visibleFieldKeys[field.Key]; shouldLog {
+			field.AddTo(enc)
+		}
+	}
+
+	// The internal zapcore.encodeError function[^0] can result in multiple fields. For example, an error type
+	// implementing fmt.Formatter results in another "errorVerbose" field, containing the stack trace if the error was
+	// created by github.com/pkg/errors including a stack[^1]. So the keys are checked again in the following loop.
+	//
+	// [^0]: https://github.com/uber-go/zap/blob/v1.27.0/zapcore/error.go#L47
+	// [^1]: https://pkg.go.dev/github.com/pkg/errors@v0.9.1#WithStack
+	visibleFields := make([]string, 0, len(visibleFieldKeys))
+	for k, v := range utils.IterateOrderedMap(enc.Fields) {
+		if _, shouldLog := visibleFieldKeys[k]; !shouldLog {
+			continue
+		}
+
+		var encodedField string
+		switch v.(type) {
+		case string, []byte, error:
+			encodedField = fmt.Sprintf("%s=%q", k, v)
+		default:
+			encodedField = fmt.Sprintf(`%s="%v"`, k, v)
+		}
+
+		visibleFields = append(visibleFields, encodedField)
+	}
+
+	if len(visibleFields) == 0 {
+		return ""
+	}
+
+	return "\t" + strings.Join(visibleFields, ", ")
 }
--- a/logging/journald_core.go
+++ b/logging/journald_core.go
@@ -67,24 +67,29 @@ func (c *journaldCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
 	}
 
 	enc := zapcore.NewMapObjectEncoder()
-	// Ensure that all field keys are valid journald field keys. If in doubt, use encodeJournaldFieldKey.
 	c.addFields(enc, fields)
 	c.addFields(enc, c.context)
 	enc.Fields["SYSLOG_IDENTIFIER"] = c.identifier
+
+	// Re-encode keys before passing them to journald. Unfortunately, this cannot be done within addFields or at another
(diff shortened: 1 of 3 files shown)


yhabteab (commented):
> An exemplary Icinga DB log message would change as follows while still containing the ICINGADB_ERROR field.
> 
> > database: Can't connect to database. Retrying
> > database: Can't connect to database. Retrying	error="dial tcp [::1]:5432: connect: connection refused"

Well, that looks quite nice and informative but wouldn't this also include a bunch of stacktraces if the error has any? That looks even scarier and disasteriouser IMHO.

```bash
Starting Icinga DB daemon (1.3.0)
Connecting to database at 'mysql://icingadb@localhost:3306/icingadb'
database: Can't connect to database. Retrying	error="dial tcp [::1]:3306: connect: connection refused", errorVerbose="dial tcp [::1]:3306: connect: connection refused\ngithub.com/icinga/icinga-go-library/database.RetryConnector.Connect.func2\n\t/Users/yhabteab/Workspace/go/icinga-go-library/database/driver.go:71\ngithub.com/icinga/icinga-go-library/retry.WithBackoff\n\t/Users/yhabteab/Workspace/go/icinga-go-library/retry/retry.go:106\ngithub.com/icinga/icinga-go-library/database.RetryConnector.Connect\n\t/Users/yhabteab/Workspace/go/icinga-go-library/database/driver.go:48\ndatabase/sql.(*DB).conn\n\t/opt/homebrew/Cellar/go/1.24.2/libexec/src/database/sql/sql.go:1431\ndatabase/sql.(*DB).PingContext.func1\n\t/opt/homebrew/Cellar/go/1.24.2/libexec/src/database/sql/sql.go:900\ndatabase/sql.(*DB).retry\n\t/opt/homebrew/Cellar/go/1.24.2/libexec/src/database/sql/sql.go:1576\ndatabase/sql.(*DB).PingContext\n\t/opt/homebrew/Cellar/go/1.24.2/libexec/src/database/sql/sql.go:899\ndatabase/sql.(*DB).Ping\n\t/opt/homebrew/Cellar/go/1.24.2/libexec/src/database/sql/sql.go:917\nmain.run\n\t/Users/yhabteab/Workspace/go/icingadb/cmd/icingadb/main.go:65\nmain.main\n\t/Users/yhabteab/Workspace/go/icingadb/cmd/icingadb/main.go:37\nruntime.main\n\t/opt/homebrew/Cellar/go/1.24.2/libexec/src/runtime/proc.go:283\nruntime.goexit\n\t/opt/homebrew/Cellar/go/1.24.2/libexec/src/runtime/asm_arm64.s:1223"
```

oxzi:
@yhabteab: How did you created this output? It contains both the `error` and the `errorVerbose` field. The latter carries the stack trace and should have not been included here. For reference, it is not part of the output I shared in the commit message/PR description.

yhabteab:
> @yhabteab: How did you created this output?

Well, obviously I don't have systemd installed on my Mac, so I had workaround this like this:

**IcingaDB:**

```diff
diff --git a/cmd/icingadb/main.go b/cmd/icingadb/main.go
index 3e2815a..b1ebc55 100644
--- a/cmd/icingadb/main.go
+++ b/cmd/icingadb/main.go
@@ -40,6 +40,7 @@ func main() {
 func run() int {
        cmd := command.New()

+       cmd.Config.Logging.Output = logging.JOURNAL
        logs, err := logging.NewLoggingFromConfig(utils.AppName(), cmd.Config.Logging)
        if err != nil {
                utils.PrintErrorThenExit(err, ExitFailure)
```

**Icinga Go Library:**

```diff
diff --git a/database/driver.go b/database/driver.go
index 5aa03aea..d83a9ed6 100644
--- a/database/driver.go
+++ b/database/driver.go
@@ -68,7 +68,7 @@ func (c RetryConnector) Connect(ctx context.Context) (driver.Conn, error) {
                                }

                                if lastErr == nil || err.Error() != lastErr.Error() {
-                                       c.logger.Warnw("Can't connect to database. Retrying", zap.Error(err))
+                                       c.logger.Warnw("Can't connect to database. Retrying", zap.Error(errors.WithStack(err)))
                                }
                        },
                        OnSuccess: func(elapsed time.Duration, attempt uint64, lastErr error) {
diff --git a/logging/journald_core.go b/logging/journald_core.go
index 263635e4..b4693fa4 100644
--- a/logging/journald_core.go
+++ b/logging/journald_core.go
@@ -6,6 +6,7 @@ import (
        "github.com/pkg/errors"
        "github.com/ssgreg/journald"
        "go.uber.org/zap/zapcore"
+       "log"
        "strings"
 )

@@ -88,6 +89,8 @@ func (c *journaldCore) Write(ent zapcore.Entry, fields []zapcore.Field) error {
        }
        message += "\t" + strings.Join(visibleFields, ", ")

+       log.Println(message)
+       return nil
        return journald.Send(message, pri, enc.Fields)
 }
```

I know that this specific reconnect error won't have any stacktraces, I just wanted to demonstrate how that would look like with an error with stacks.

logging/journald_core.go
```diff
@@ -124,3 +132,50 @@ func encodeJournaldFieldKey(key string) string {
...
+	// capacity of twice the fields length to minimize the chance of a resize.
+	//
+	// [^0]: https://github.com/uber-go/zap/blob/v1.27.0/zapcore/error.go#L47
+	// [^1]: https://pkg.go.dev/github.com/pkg/errors@v0.9.1#WithStack
+	visibleFields := make([]string, 0, 2*len(enc.Fields))
+	for k, v := range utils.IterateOrderedMap(enc.Fields) {
+		if _, shouldLog := visibleFieldKeys[k]; !shouldLog {
+			continue
+		}
+		visibleFields = append(visibleFields, fmt.Sprintf("%s=%q", k, v))
```
yhabteab, line 173: Using the `%q` as a format specifier here will only work for string values, for all other types like for example `int` will result into such weird output:
```bash
2025/05/12 09:22:04 Starting Icinga DB daemon (1.3.0)
2025/05/12 09:22:04 Connecting to database at 'mysql://icingadb@localhost:3306/icingadb'
2025/05/12 09:22:04 database: Can't connect to database. Retrying	error="dial tcp [::1]:3306: connect: connection refused", int='d'
```

oxzi: Good catch. So far, I have only though about strings and errors, which are kinda strings under the hood. I have added a bit of generalization to distinct between `%q`-aware types and others. However, I would not make this too complicated as for the moment only `error` fields are relevant. Please take another look.

yhabteab: Do all the values have to be quoted with `"..."`? Otherwise, why not just append the values using the `%v` verb in the default case?

logging/journald_core.go
```diff
@@ -124,3 +132,61 @@ func encodeJournaldFieldKey(key string) string {
...
+		switch v.(type) {
+		case rune, string, []byte, error:
+			fieldVerb = "%q"
+			// %q already enclosures the value
+		default:
+			fieldVerb = "%v"
+			fieldEnclosure = `"`
+		}
+
+		visibleFields = append(visibleFields, fmt.Sprintf("%s=%s"+fieldVerb+"%s", k, fieldEnclosure, v, fieldEnclosure))
```
julianbrost, line 184: Don't you end up with exactly two different format strings here in the end? Wouldn't it be much simpler to just specify both of them explicitly?

logging/journald_core.go
oxzi: Of course they don't have to be quoted. However, I found it easier to read if each value is encapsulated, especially if multiple fields are there.

logging/journald_core.go
oxzi: I have merged them into one format string. Please take another look.

julianbrost: I was referring to the whole format string, just having the whole string there is just easier to read, even if you repeat the `%s=`.

I'd even go so far and move the `fmt.Sprintf()` into the different cases as well (that should make it easier for linters):

```go
		var encodedField string
		switch v.(type) {
		case rune, string, []byte, error:
			encodedField = fmt.Sprintf("%s=%q", k, v)
		default:
			encodedField = fmt.Sprintf(`%s="%v"`, k, v)
		}

		visibleFields = append(visibleFields, encodedField)
```

logging/journald_core.go
```diff
@@ -124,3 +132,59 @@ func encodeJournaldFieldKey(key string) string {
...
+	visibleFields := make([]string, 0, 2*len(enc.Fields))
+	for k, v := range utils.IterateOrderedMap(enc.Fields) {
+		if _, shouldLog := visibleFieldKeys[k]; !shouldLog {
+			continue
+		}
+
+		var fieldVerb string
+		switch v.(type) {
+		case rune, string, []byte, error:
+			fieldVerb = `%q`
```
julianbrost, line 177: I doubt having `rune` in here is a good idea as it's just an [alias for `int32`](https://pkg.go.dev/builtin#rune). I think we're more likely to log an `int32` than a `rune` anywhere. That would also be worth a test case.

(I'm glad that Go is just as broken as C++ in this respect.[])

logging/journald_core.go
oxzi: That even makes it easier. I have pushed a new version.

logging/journald_core.go
oxzi: Indeed. I have removed `rune` from the `%q` verb case. So far, I cannot think of a case where we have logged a rune[^1].

[^1]: I am not asking for a proof by contradiction :upside_down_face:

logging/journald_core.go
yhabteab: Fine!

yhabteab (approved):
LFTM!
```bash
2025/05/14 12:46:02 Starting Icinga DB daemon (1.3.0)
2025/05/14 12:46:02 Connecting to database at 'mysql://icingadb@localhost:3306/icingadb'
2025/05/14 12:46:02 database: Can't connect to database. Retrying	attempt="1", elapsed="3.8735ms", error="dial tcp [::1]:3306: connect: connection refused"
```

oxzi:
After discussion this PR and a more border change with @lippserd, we realized that at the moment the stack traces (`errorVerbose`) were missing from the journald logs while being visible when using the console output.

After a short debugging session, the culprit was found: the key name sanitation was called for the original name, but was missing for derived names, such as `SOMETHING_ERRORVerbose`.

I have decided to add the patch to this PR since these changes are somewhat related.

oxzi:
@lippserd: This worked for me.

These lines...

```go
c.logger.Warnw("Can't connect to database. Retrying", zap.Error(errors.Wrap(err, "zap oops")))
c.logger.Warnw("Can't connect to database. Retrying", logging.Error(errors.Wrap(err, "igl oops")))
```

...produced resulted in the following output:

```
May 15 17:01:56 nwt14 icingadb[566989]: database: Can't connect to database. Retrying        error="zap oops: dial tcp [::1]:3306: connect: connection refused"
May 15 17:01:56 nwt14 icingadb[566989]: database: Can't connect to database. Retrying        error="igl oops: dial tcp [::1]:3306: connect: connection refused"
```

When viewing journalctl with verbose output, the `ICINGADB_ERROR_VERBOSE` field is only present for the "zap oops" entry.

oxzi:
@lippserd: Thanks for rebasing.

PR #127, opened 2025-04-28
Import Icinga Notifications `internal/utils`

oxzi:
This PR imports the `internal/utils` package from Icinga Notifications into `utils` for general purpose functions and into `database` for database-related utility functions. The commit history was kept, but messages were slightly altered since the original messages often referred changes outside the `internal/utils` package.

My motivation for this merge was to use a utility function already implemented, but residing in the wrong project's utility package. Since the IGL is the place for common code, I would like to get rid of utility packages outside this repository.

<details><summary>Steps performed to create this PR</summary>
<p>

## Create Patches

Within the `icinga-notifications` repository.

```sh
$ cd internal/utils
$ git format-patch --relative --root -- ./
0001-Introduce-utils-package.patch
0002-Move-helper-function-to-utils-package.patch
0003-Implement-incremental-config-updates-for-rules.patch
0004-Add-utils-ToDBInt-function.patch
0005-Use-sqlx.Tx-for-event-processing.patch
0006-Use-ctx-everywhere-cache-object-only-when-the-tx-suc.patch
0007-Trigger-time-based-escalations.patch
0008-Rename-and-simplify-restoreAttrsFor-function.patch
0009-Make-sure-to-release-semaphore-in-utils-ForEachRow.patch
0010-utils.IterateOrderedMap-for-plugin.FormatMessage.patch
0011-Switch-to-icinga-go-library.patch
0012-Drop-unused-Remove-utils-functions.patch
0013-Restore-all-muted-objects-without-an-incident-on-dae.patch
0014-Make-logging-more-consistent.patch
0015-Use-utils.IterateOrderedMap-as-intended-via-for-rang.patch
$ cd -
$ mv internal/utils/*.patch .
$ ls *.patch
```

## Apply Patches

Within this repository.

First, create a `githook(5)` to verify each applied patch works.
Plot twist: each patch will fail now.

```sh
$ cat .git/hooks/pre-applypatch
#!/usr/bin/env bash
#
# An example hook script to verify what is about to be committed
# by applypatch from an e-mail message.
#
# The hook should exit with non-zero status after issuing an
# appropriate message if it wants to stop the commit.
#
# To enable this hook, rename this file to "pre-applypatch".

. git-sh-setup
precommit="$(git rev-parse --git-path hooks/pre-commit)"
test -x "$precommit" && exec "$precommit" ${1+"$@"}

go build
go test -race -count=1 ./...
```

Start applying patches:

```sh
$ git am -3 --directory utils/ ../icinga-notifications/*.patch
```

Each patch now requires some manual intervention, e.g., making it compile or manually move some code to `./database/util.go`.

If the current patch works, build it manually once and continue afterwards.

```sh
$ git add utils/ database/ && git am --continue
```

The `0011-Switch-to-icinga-go-library.patch` patch was skipped since it does not made sense in this context.

## Verifying Methods

To make sure that nothing was lost, I have manually diffed each function.

```diff
@@ -1,5 +1,5 @@
 // BuildInsertStmtWithout builds an insert stmt without the provided column.
-func BuildInsertStmtWithout(db *database.DB, into interface{}, withoutColumn string) string {
+func BuildInsertStmtWithout(db *DB, into interface{}, withoutColumn string) string {
        columns := db.BuildColumns(into)
        for i, column := range columns {
                if column == withoutColumn {
@@ -11,7 +11,7 @@

        return fmt.Sprintf(
                `INSERT INTO "%s" ("%s") VALUES (%s)`,
-               database.TableName(into), strings.Join(columns, `", "`),
+               TableName(into), strings.Join(columns, `", "`),
                fmt.Sprintf(":%s", strings.Join(columns, ", :")),
        )
 }
```

```diff
@@ -3,7 +3,7 @@
 // A new transaction is started on db which is then passed to fn. After fn returns, the transaction is
 // committed unless an error was returned. If fn returns an error, that error is returned, otherwise an
 // error is returned if a database operation fails.
-func RunInTx(ctx context.Context, db *database.DB, fn func(tx *sqlx.Tx) error) error {
+func RunInTx(ctx context.Context, db *DB, fn func(tx *sqlx.Tx) error) error {
        tx, err := db.BeginTxx(ctx, nil)
        if err != nil {
                return err
```

```diff
@@ -1,7 +1,7 @@
 // InsertAndFetchId executes the given query and fetches the last inserted ID.
 func InsertAndFetchId(ctx context.Context, tx *sqlx.Tx, stmt string, args any) (int64, error) {
        var lastInsertId int64
-       if tx.DriverName() == database.PostgreSQL {
+       if tx.DriverName() == PostgreSQL {
                preparedStmt, err := tx.PrepareNamedContext(ctx, stmt+" RETURNING id")
                if err != nil {
                        return 0, err
```

```diff
@@ -1,7 +1,7 @@
 // ExecAndApply applies the provided restoreFunc callback for each successfully retrieved row of the specified type.
 // Returns error on any database failure or fails to acquire the table semaphore.
-func ExecAndApply[Row any](ctx context.Context, db *database.DB, stmt string, args []interface{}, restoreFunc func(*Row)) error {
-       table := database.TableName(new(Row))
+func ExecAndApply[Row any](ctx context.Context, db *DB, stmt string, args []interface{}, restoreFunc func(*Row)) error {
+       table := TableName(new(Row))
        sem := db.GetSemaphoreForTable(table)
        if err := sem.Acquire(ctx, 1); err != nil {
                return errors.Wrapf(err, "cannot acquire semaphore for table %q", table)
```

```diff
@@ -1,7 +1,7 @@
 // ForEachRow applies the provided restoreFunc callback for each successfully retrieved row of the specified type.
 // It will bulk SELECT the data from the database scoped to the specified ids and scans into the provided Row type.
 // Returns error on any database failure or fails to acquire the table semaphore.
-func ForEachRow[Row, Id any](ctx context.Context, db *database.DB, idColumn string, ids []Id, restoreFunc func(*Row)) error {
+func ForEachRow[Row, Id any](ctx context.Context, db *DB, idColumn string, ids []Id, restoreFunc func(*Row)) error {
        subject := new(Row)
        query := fmt.Sprintf("%s WHERE %q IN (?)", db.BuildSelectStmt(subject, subject), idColumn)
        stmt, args, err := sqlx.In(query, ids)
```

No diff for `ToDBString`, `ToDBInt`, and `IterateOrderedMap`.

## Reword Commit Messages

Since most commit messages referred a certain change and the changes in the `util.go` file was a drive-by change, I have reworded each commit message and linked the original commit on GitHub.

## Cleanup

As further specified in <https://github.com/Icinga/icinga-go-library/pull/127#issuecomment-2838713035>, some functions and commits were dropped again.

</p>
</details>

### DIFF
--- a/database/utils.go
+++ b/database/utils.go
@@ -10,6 +10,7 @@ import (
 	"github.com/icinga/icinga-go-library/types"
 	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
+	"slices"
 	"strings"
 )
 
@@ -82,15 +83,11 @@ func InsertObtainID(ctx context.Context, conn TxOrDB, stmt string, arg any) (int
 	return resultID, nil
 }
 
-// BuildInsertStmtWithout builds an insert stmt without the provided column.
-func BuildInsertStmtWithout(db *DB, into interface{}, withoutColumn string) string {
-	columns := db.BuildColumns(into)
-	for i, column := range columns {
-		if column == withoutColumn {
-			columns = append(columns[:i], columns[i+1:]...)
-			break
-		}
-	}
+// BuildInsertStmtWithout builds an insert stmt without the provided columns.
+func BuildInsertStmtWithout(db *DB, into interface{}, withoutColumns ...string) string {
+	columns := slices.DeleteFunc(
+		db.BuildColumns(into),
+		func(column string) bool { return slices.Contains(withoutColumns, column) })
 
 	return fmt.Sprintf(
 		`INSERT INTO "%s" ("%s") VALUES (%s)`,
--- a/database/utils.go
+++ b/database/utils.go
@@ -10,6 +10,7 @@ import (
 	"github.com/icinga/icinga-go-library/types"
 	"github.com/jmoiron/sqlx"
 	"github.com/pkg/errors"
+	"strings"
 )
 
 // CantPerformQuery wraps the given error with the specified query that cannot be executed.
@@ -79,6 +80,23 @@ func InsertObtainID(ctx context.Context, conn TxOrDB, stmt string, arg any) (int
 	}
 
 	return resultID, nil
+}
+
+// BuildInsertStmtWithout builds an insert stmt without the provided column.
+func BuildInsertStmtWithout(db *DB, into interface{}, withoutColumn string) string {
+	columns := db.BuildColumns(into)
+	for i, column := range columns {
+		if column == withoutColumn {
+			columns = append(columns[:i], columns[i+1:]...)
+			break
+		}
+	}
+
+	return fmt.Sprintf(
+		`INSERT INTO "%s" ("%s") VALUES (%s)`,
+		TableName(into), strings.Join(columns, `", "`),
+		fmt.Sprintf(":%s", strings.Join(columns, ", :")),
+	)
 }
 
 // unsafeSetSessionVariableIfExists sets the given MySQL/MariaDB system variable for the specified database session.
--- a/types/int.go
+++ b/types/int.go
@@ -14,6 +14,29 @@ type Int struct {
 	sql.NullInt64
 }
 
+// TransformZeroIntToNull transforms a valid Int carrying a zero value to a SQL NULL.
+func TransformZeroIntToNull(i *Int) {
+	if i.Valid && i.Int64 == 0 {
+		i.Valid = false
+	}
+}
+
+// MakeInt constructs a new Int.
+//
+// Multiple transformer functions can be given, each transforming the generated Int, e.g., TransformZeroIntToNull.
+func MakeInt(in int64, transformers ...func(*Int)) Int {
+	i := Int{sql.NullInt64{
+		Int64: in,
+		Valid: true,
+	}}
+
+	for _, transformer := range transformers {
+		transformer(&i)
+	}
+
+	return i
+}
+
 // MarshalJSON implements the json.Marshaler interface.
 // Supports JSON null.
 func (i Int) MarshalJSON() ([]byte, error) {
--- a/types/string.go
+++ b/types/string.go
@@ -14,12 +14,27 @@ type String struct {
 	sql.NullString
 }
 
-// MakeString constructs a new non-NULL String from s.
-func MakeString(s string) String {
-	return String{sql.NullString{
-		String: s,
+// TransformEmptyStringToNull transforms a valid String carrying an empty text to a SQL NULL.
+func TransformEmptyStringToNull(s *String) {
+	if s.Valid && s.String == "" {
+		s.Valid = false
+	}
+}
+
+// MakeString constructs a new String.
+//
+// Multiple transformer functions can be given, each transforming the generated String, e.g., TransformEmptyStringToNull.
+func MakeString(in string, transformers ...func(*String)) String {
+	s := String{sql.NullString{
+		String: in,
 		Valid:  true,
 	}}
+
+	for _, transformer := range transformers {
+		transformer(&s)
+	}
+
+	return s
 }
 
 // MarshalJSON implements the json.Marshaler interface.
--- a/types/int_test.go
+++ b/types/int_test.go
@@ -6,6 +6,49 @@ import (
 	"testing"
 )
 
+func TestMakeInt(t *testing.T) {
+	subtests := []struct {
+		name         string
+		input        int64
+		transformers []func(*Int)
+		output       sql.NullInt64
+	}{
+		{
+			name:   "zero",
+			input:  0,
+			output: sql.NullInt64{Int64: 0, Valid: true},
+		},
+		{
+			name:   "positive",
+			input:  1,
+			output: sql.NullInt64{Int64: 1, Valid: true},
+		},
+		{
+			name:   "negative",
+			input:  -1,
+			output: sql.NullInt64{Int64: -1, Valid: true},
+		},
+		{
+			name:         "zero-transform-zero-to-null",
+			input:        0,
+			transformers: []func(*Int){TransformZeroIntToNull},
+			output:       sql.NullInt64{Valid: false},
+		},
+		{
+			name:         "positive-transform-zero-to-null",
+			input:        1,
+			transformers: []func(*Int){TransformZeroIntToNull},
+			output:       sql.NullInt64{Int64: 1, Valid: true},
+		},
+	}
+
+	for _, st := range subtests {
+		t.Run(st.name, func(t *testing.T) {
+			require.Equal(t, Int{NullInt64: st.output}, MakeInt(st.input, st.transformers...))
+		})
+	}
+}
+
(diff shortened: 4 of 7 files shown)


lippserd (changes requested):

database/utils.go
```diff
@@ -81,6 +83,90 @@ func InsertObtainID(ctx context.Context, conn TxOrDB, stmt string, arg any) (int
...
+	query := fmt.Sprintf("%s WHERE %q IN (?)", db.BuildSelectStmt(subject, subject), idColumn)
+	stmt, args, err := sqlx.In(query, ids)
+	if err != nil {
+		return errors.Wrapf(err, "cannot build placeholders for %q", query)
+	}
+
+	return ExecAndApply(ctx, db, stmt, args, restoreFunc)
+}
+
+// ToDBString transforms the given string to types.String.
```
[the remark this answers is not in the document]
oxzi: Valid points. I have followed the suggestion and created methods to `types.String` and `types.int` following the builder pattern. Please take another look.

database/utils.go
```diff
@@ -81,6 +83,90 @@ func InsertObtainID(ctx context.Context, conn TxOrDB, stmt string, arg any) (int
 	return resultID, nil
 }
 
+// BuildInsertStmtWithout builds an insert stmt without the provided column.
+func BuildInsertStmtWithout(db *DB, into interface{}, withoutColumn string) string {
```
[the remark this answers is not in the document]
oxzi: Good idea. Doing so does even keep the API compatible. Please take another look.

database/utils.go
```diff
@@ -81,6 +83,90 @@ func InsertObtainID(ctx context.Context, conn TxOrDB, stmt string, arg any) (int
...
+	columns := db.BuildColumns(into)
+	for i, column := range columns {
+		if column == withoutColumn {
+			// Event id is auto incremented, so just erase it from our insert columns
```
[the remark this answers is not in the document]
oxzi: Done. Please take another look.

database/utils.go
```diff
@@ -81,6 +83,90 @@ func InsertObtainID(ctx context.Context, conn TxOrDB, stmt string, arg any) (int
...
+			break
+		}
+	}
+
+	return fmt.Sprintf(
+		`INSERT INTO "%s" ("%s") VALUES (%s)`,
+		TableName(into), strings.Join(columns, `", "`),
+		fmt.Sprintf(":%s", strings.Join(columns, ", :")),
+	)
+
```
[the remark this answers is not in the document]
oxzi: Done. Please take another look.

database/utils.go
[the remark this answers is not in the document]
lippserd: I prefer the type to be fully defined at creation rather than relying on functions that might produce different results than the originally constructed type.

Julian's points are valid, though. I see two options:
1. Do not implement this here.
2. Ensure types are fixed upon creation, e.g., using `types.NewString(s, types.AlwaysValid())` or `types.NewString(s, types.ZeroIsInvalid())`.

The second option would also allow for the creation of custom reusable functions to define when a type is considered valid.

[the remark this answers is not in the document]
oxzi: I followed the second option with an updated `Make{Int,String}` function. Feel free to take another look.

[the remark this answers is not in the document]
oxzi: @julianbrost commented that the name validator may be misleading. His suggestion was to change this to, for example, a transformer and pass a pointer. Such a transformer could change the value to `Valid: false`.

database/utils.go
```diff
@@ -81,6 +83,64 @@ func InsertObtainID(ctx context.Context, conn TxOrDB, stmt string, arg any) (int
...
+	return fmt.Sprintf(
+		`INSERT INTO "%s" ("%s") VALUES (%s)`,
+		TableName(into), strings.Join(columns, `", "`),
+		fmt.Sprintf(":%s", strings.Join(columns, ", :")),
+	)
+}
+
+// ExecAndApply applies the provided restoreFunc callback for each successfully retrieved row of the specified type.
+// Returns error on any database failure or fails to acquire the table semaphore.
+func ExecAndApply[Row any](ctx context.Context, db *DB, stmt string, args []interface{}, restoreFunc func(*Row)) error {
```
lippserd, line 101: Please remove from this PR.

database/utils.go
```diff
@@ -81,6 +83,64 @@ func InsertObtainID(ctx context.Context, conn TxOrDB, stmt string, arg any) (int
...
+		restoreFunc(row)
+	}
+
+	return rows.Err()
+}
+
+// ForEachRow applies the provided restoreFunc callback for each successfully retrieved row of the specified type.
+// It will bulk SELECT the data from the database scoped to the specified ids and scans into the provided Row type.
+// Returns error on any database failure or fails to acquire the table semaphore.
+func ForEachRow[Row, Id any](ctx context.Context, db *DB, idColumn string, ids []Id, restoreFunc func(*Row)) error {
```
lippserd, line 133: Please remove from this PR.

types/int.go
```diff
@@ -14,6 +14,28 @@ type Int struct {
...
+func MakeInt(in int64, transformers ...func(*Int)) Int {
+	i := Int{sql.NullInt64{
+		Int64: in,
+		Valid: true,
+	}}
+
+	for _, transformer := range transformers {
+		transformer(&i)
+	}
+	return i
```
lippserd, line 36: Please add newline.

types/string.go
```diff
@@ -14,12 +14,26 @@ type String struct {
...
+func MakeString(in string, transformers ...func(*String)) String {
+	s := String{sql.NullString{
+		String: in,
 		Valid:  true,
 	}}
+
+	for _, transformer := range transformers {
+		transformer(&s)
+	}
+	return s
```
lippserd, line 36: Please add newline.

types/int.go
```diff
@@ -14,6 +14,28 @@ type Int struct {
 	sql.NullInt64
 }
 
+// TransformIntZeroToNull transforms a valid Int carrying a zero value to a SQL NULL.
+func TransformIntZeroToNull(i *Int) {
```
lippserd, line 18: I would rename this to `TransformZeroIntToNull` as it is also `EmptyString` and not `StringEmpty`. I could also imagine that we could introduce a generic version `TransformZeroValueToNull` later.

lippserd (changes requested):

database/utils.go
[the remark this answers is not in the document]
lippserd: No, I'm fine with that. I left a review with some things.

database/utils.go
oxzi: Removed.

types/int.go
oxzi: Done.

types/string.go
oxzi: Done.

types/int.go
oxzi: Renamed.

lippserd (approved):

PR #145, opened 2025-08-06
Add a common notifications client & its utilities

yhabteab:
This PR introduces a new `notifications` package to our Icinga Go library. The package is designed to handle common
Icinga Notifications related tasks, that can be used by any Icinga Notifications source. It provides a set of functions and types that simplify and centralize the interaction with Icinga Notifications API. Within the root of this package, you will find the following files:

- `client.go`: This file contains the actual client implementation with a custom HTTP round tripper that handles
  the necessary authentication before sending the final request to the Icinga API.
- `config.go`: This file defines the configuration structure that every Icinga Notifications source is required to provide.
- `xhttp_headers.go`: This file contains our custom HTTP header constants only that should be used by the Icinga
  Notifications sources to ensure consistency and compatibility with the Icinga API. Currently, it includes only
  `X-Icinga-Rules-Version` header, which is used to specify the version of the rules that are being used by the source and
  `X-Icinga-Rules-Id` header, which is used to specify the unique identifier of the rules that are being used by the source.
  Of course, in order to be consistent with the HTTP header naming conventions in Icinga Web 2, both custom headers
  are prefixed with `X-Icinga-`.

The `event` package within the `notifications` package contains common event related types and utilities that should
be used by any Icinga Notifications source as well as by the Icinga Notifications daemon itself.

The change list appears to be quite big, but don't be fooled by it :). Most of them are just unit tests and autogenerated
code. The actual code changes are quite minimal and focused on the new functionality described above.

Oops, did I say autogenerated code? Yes, I did ;)! The `notifications` package makes use of the Go [`stringer`](https://pkg.go.dev/golang.org/x/tools/cmd/stringer) tool to implement the `String()` method for the `event.Type` and `event.Severity` types in an efficient manner. So, don't be surprised to see the autogenerated files in the `event` package.

Oh, and one more thing: most of the work was done by Alvar in the Icinga DB repository, so I just had to outsource it to the Icinga Go library with some minor adjustments. So, credit goes to him for the majority of the work.


julianbrost:
I'm not sure if this is the right place to add this. The reasoning probably is similar to https://github.com/Icinga/icinga-notifications/issues/304. One thing to consider here: do we expect external projects using this eventually? In this case, this would create dependency updates for them even though if nothing relevant to them changed. So I think having a separate repository with independent versioning for this is also worth considering.

lippserd:
> I'm not sure if this is the right place to add this. The reasoning probably is similar to [Icinga/icinga-notifications#304](https://github.com/Icinga/icinga-notifications/issues/304). One thing to consider here: do we expect external projects using this eventually? In this case, this would create dependency updates for them even though if nothing relevant to them changed. So I think having a separate repository with independent versioning for this is also worth considering.

I like this idea. How about a MIT licensed `go-icinga-notifications` project with this code, and the channel base?

notifications/source/client.go
```diff
@@ -0,0 +1,196 @@
...
+// It is used to indicate that there are no rules available for a given source.
+const EmptyRulesVersion = "0x0"
+
+// RulesInfo holds information about the event rules for a specific source.
+type RulesInfo struct {
+	Version string             // Version of the event rules fetched from the API.
+	Rules   map[int64]RuleResp // Rules is a map of rule IDs to their corresponding RuleResp objects.
+}
+
+// Iter returns an iterator over the rules in the SourceRulesInfo.
```
oxzi, line 179:
```suggestion
// Iter returns an iterator over the rules in the RulesInfo.
```

yhabteab: This code won't last long though ;)! When sources start to manage their own event rules config by their own, this will become obsolete.

yhabteab:
Changed the tabular test cases to use `testutils.TestCase` instead of a custom one!

oxzi:
Removed the custom HTTP headers and moved the rules to the Event struct to address https://github.com/Icinga/icinga-notifications/pull/324#discussion_r2313547798.

notifications/source/client.go
```diff
@@ -0,0 +1,187 @@
...
+
+	"github.com/icinga/icinga-go-library/notifications/event"
+	"github.com/pkg/errors"
+)
+
+// ErrRulesOutdated implies that the rules version between Icinga DB and Icinga Notifications mismatches.
+var ErrRulesOutdated = fmt.Errorf("rules version is outdated")
+
+// BasicAuthTransport is an http.RoundTripper that adds basic authentication and a User-Agent header to HTTP requests.
+type BasicAuthTransport struct {
```
julianbrost, line 22: This type is only used internally in the implementation here, or this there actually a need to export it from the package?

notifications/source/client.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+// JoinIcingaWeb2Path constructs a URL by joining the Icinga Web 2 base URL with the provided relative URL.
+//
+// It is used to convert any relative URL into an absolute URL that points to the Icinga Web 2 instance.
+// A relative URL like "/icingadb/host" is transformed to e.g. "https://icinga.example.com/icingaweb2/icingadb/host"
+// after passing through this method, assuming the Icinga Web 2 base URL is "https://icinga.example.com/icingaweb2".
+func (c *Client) JoinIcingaWeb2Path(relativePath string) *url.URL {
+	return c.IcingaWebBaseUrl.JoinPath(relativePath)
+}
```
julianbrost, line 158: > A relative URL like "/icingadb/host" is transformed to e.g. "https://icinga.example.com/icingaweb2/icingadb/host"

If you actually had a link to that on `https://icinga.example.com/icingaweb2`, it would resolve to `https://icinga.example.com/icingadb/host`, so the behavior of the function doesn't really match how releative URLs are treated, so referring to that leaves a bit of room for confusion IMHO.

I'd rather suggest wording like "Interprets the given path to be relative to the Icinga Web base URL and returns an absolute URL for it."

However, on a more general note: How much sense does having this here actually make? I mean `NewClient()` requires the source to pass a URL as part of `Config` just to provide this one line function? Couldn't that also be trivially handles within the source? Or, what would probably even make more sense: shouldn't we just change the definition of that can be sent in an event so that a source can just send a path relative to the Icinga Web installation root URL and the notification daemon takes care of the rest, i.e. only that needs to know where Icinga Web is living, not every single source.

notifications/source/client.go
```diff
@@ -0,0 +1,187 @@
...
+// It holds the configuration for the API endpoint and the HTTP client used to make requests.
+type Client struct {
+	cfg Config // cfg holds base API endpoint URL and authentication details.
+
+	client http.Client // HTTP client used for making requests to the Icinga Notifications API.
+
+	IcingaWebBaseUrl *url.URL // IcingaWebBaseUrl holds the base URL for Icinga Web 2.
+	Endpoints        struct {
+		ProcessEvent string // ProcessEvent holds the URL for the process event endpoint.
+	}
```
julianbrost, line 50: I currently don't see a point in having that extra struct. I know that there was a second URL in there earlier, so it was probably intended for some grouping within the `Client` struct, but now it just looks strange.

Additionally, I don't see why this should be exported.

notifications/source/client.go
```diff
@@ -0,0 +1,187 @@
...
+			},
+		},
+	}
+
+	baseUrl, err := url.Parse(cfg.ApiBaseUrl)
+	if err != nil {
+		return nil, errors.Wrap(err, "unable to parse API base URL")
+	}
+
+	client.Endpoints.ProcessEvent = baseUrl.ResolveReference(&url.URL{Path: "/process-event"}).String()
```
julianbrost, line 78: Why does it use the more complex `ResolveReference()` over `JoinPath()`? This just makes it possible to have strange configuration like `ApiBaseUrl = "http://localhost:1234/what-am-i-even-doing"` and I don't see a good reason why that should work.

notifications/source/client.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, errors.Wrap(err, "unable to parse Icinga Web 2 base URL")
+	}
+
+	return client, nil
+}
+
+// ProcessEvent submits an event to the Icinga Notifications /process-event API endpoint.
+//
+// It serializes the event into JSON and sends it as a POST request to the process event endpoint. In most cases, the
+// Event.RulesVersion and Event.RuleIds must be set.
```
julianbrost, line 91: > In most cases

This isn't very helpful if the whole doc comment doesn't mention the rare other case.

Isn't the other case the first event sent by a source, so the source will typically even need to do that, so that should be mentioned.

notifications/source/client.go
```diff
@@ -0,0 +1,187 @@
...
+// It is used to convert any relative URL into an absolute URL that points to the Icinga Web 2 instance.
+// A relative URL like "/icingadb/host" is transformed to e.g. "https://icinga.example.com/icingaweb2/icingadb/host"
+// after passing through this method, assuming the Icinga Web 2 base URL is "https://icinga.example.com/icingaweb2".
+func (c *Client) JoinIcingaWeb2Path(relativePath string) *url.URL {
+	return c.IcingaWebBaseUrl.JoinPath(relativePath)
+}
+
+// EmptyRulesVersion is a constant representing the version of the rules when no rules are present.
+// It is used to indicate that there are no rules available for a given source.
+const EmptyRulesVersion = "0x0"
```
julianbrost, line 162: Do we need this? We shouldn't need to treat this specially on the source side, that should just see any opaque rule version, no matter if this describes the empty rule set or not. (We could treat the empty ruleset special on the daemon side though because it's empty, we know that the result of evaluation it will be that no rules match, so in that case, the event can be accepted without having the check the rule version.)

notifications/event/event.go
```diff
@@ -0,0 +1,45 @@
...
+	// MuteReason provides a reason for muting the object if Mute is set to true.
+	//
+	// Setting this field to an empty string while Mute is true will cause the request to fail,
+	// as Icinga Notifications requires a reason for muting an object. Otherwise, it will be omitted
+	// from the encoded JSON.
+	MuteReason string `json:"mute_reason,omitempty"`
+
+	// RulesVersion and RuleIds are the source rules matching for this Event.
+	RulesVersion string  `json:"rules_version"`
+	RuleIds      []int64 `json:"rule_ids"`
```
julianbrost, line 44: If possible without too much work, we could switch to `[]string` here, allowing us to switch to another identifier (like for example a UUID) in the future. Though `int64` will probably be good enough, so not too important to have this changed, we could just make the actual data type used in the schema opaque to the source.

julianbrost:
Oh and the PR description is quite outdated by now. It mentions headers that are gone for example.

notifications/source/client.go
oxzi: Done.

notifications/source/client.go
oxzi: Dunno. Changed.

notifications/source/client.go
oxzi: Reworded.

notifications/source/client.go
oxzi: No, this was mostly an artifact from prior versions of this PR. Removed.

notifications/event/event.go
```diff
@@ -0,0 +1,51 @@
...
+	// It is a map of string keys to string values, allowing for flexible tagging of events if the event
+	// name alone is not sufficient to identify the object. In the case of using Icinga DB as a source, the
+	// tags will typically look like this:
+	// For hosts: {"host": "host_name"} and for services: {"host": "host_name", "service": "service_name"}.
+	Tags map[string]string `json:"tags"`
+
+	Type     Type     `json:"type"`               // Type indicates the type of the event (see Type for possible values).
+	Severity Severity `json:"severity,omitempty"` // The severity of the event (see Severity for possible values).
+	Username string   `json:"username"`           // Username is the name of the user who triggered the event.
+	Message  string   `json:"message"`            // Message is a human-readable message describing the event.
```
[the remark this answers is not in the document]
oxzi: Done.

notifications/source/client.go
```diff
@@ -0,0 +1,168 @@
...
+
+	return client, nil
+}
+
+// ProcessEvent submits an event to the Icinga Notifications /process-event API endpoint.
+//
+// It serializes the event into JSON and sends it as a POST request to the process event endpoint.
+//
+// Event.RulesVersion and Event.RuleIds must be set. When no information is available, set them to an empty string and
+// an empty []int64, respectively.
```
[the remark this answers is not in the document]
oxzi: Changed, thanks. The old message was both incorrect `[]string` instead of `[]int64` and, IIRC, there was one reason to do so in the meantime.

notifications/event/event.go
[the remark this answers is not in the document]
yhabteab: > Those comments should use the standard doc format (i.e. be in the line before the member).

Why do you think this isn't a [standard doc format](https://tip.golang.org/doc/comment#type)?

notifications/source/client.go
```diff
@@ -0,0 +1,168 @@
...
+				break
+			}
+		}
+	}
+}
+
+// RuleResp represents a response object for a rule in the Icinga Notifications API.
+type RuleResp struct {
+	Id               string // Id is the unique identifier of the rule.
+	Name             string // Name is the name of the rule.
```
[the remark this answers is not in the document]
oxzi: Valid point. I have removed the `RulesResp` type. Rules are now just represented by a rule ID to filter expression mapping in the `RulesInfo` struct.

notifications/event/event.go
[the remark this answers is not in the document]
julianbrost: Well okay, for structs, godoc just shows the full definition anyway ([see `http.Client` for example](https://pkg.go.dev/net/http#Client)). I Just found it look strange that part of the struct had the (I'd say more common) comments above, with these four doing it differently.

PR #148, opened 2025-08-13
redis.NewClientFromConfig(): remove obsolete workaround

Al2Klimov:
The underlying Redis client library doesn't blindly use eventually broken connections from the pool anymore. Hence, we don't have to re-try commands more often than the pool size.


Al2Klimov:
# Test

Currently we're using github.com/redis/go-redis/v9 v9.11.0 which changes the behavior of the test program from https://github.com/redis/go-redis/issues/1737 OP.

Original: "use of closed network connection"

v9.11.0: "Dialer called"

Al2Klimov:
Confirmed by https://github.com/redis/go-redis/issues/1737#issuecomment-3200385648.

PR #199, opened 2026-04-30
notifications: drop obsolete event fields & introduce new ones

yhabteab:
> [!WARNING]
>
> This PR is not yet part of the main branch since it introduces a breaking change to the notifications package and all its users must be fixed first before merging this into the main branch (See the failing GHAs below).

This PR drops the obsolete `extra_tags`, `rule_ids`, and `rules_version` fields from the notification `Event` struct as they are no longer used by Icinga Notifications since https://github.com/Icinga/icinga-notifications/pull/418. It also introduces two new fields `complete_relations` and `relations` to the `Event` struct, which are going to be used by the new Notifications rule evaluation engine to determine which rules match a given event. Additionally, it adds a new HTTP header `X-Icinga-Reject-If-Relations-Incomplete` to be used by the Notifications API to indicate whether the client supports attribute negotiation. The name of the header is not final and is open for discussion.

### DIFF
--- a/notifications/event/event.go
+++ b/notifications/event/event.go
@@ -27,9 +27,6 @@ type Event struct {
 	// For hosts: {"host": "host_name"} and for services: {"host": "host_name", "service": "service_name"}.
 	Tags map[string]string `json:"tags"`
 
-	// ExtraTags supplement Tags, for example with host or service groups for an Icinga DB source.
-	ExtraTags map[string]string `json:"extra_tags"`
-
 	// Type indicates the type of the event.
 	Type Type `json:"type"`
 	// Severity of the event.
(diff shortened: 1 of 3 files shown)


oxzi:
While I am generally okay with this change, the failing compatibility CI jobs underline a certain question: Do we want to merge this directly into `main` and temporarily breaking compatibility with the downstream consumers or do we want to work with a feature branch to merge things into before merging it all back to `main`?

yhabteab:
Just in case we need to release IGL in the meantime without such changes, we can also use a separate branch till all the consumers are adapted to these changes as well.

yhabteab:
I've changed the base branch now.

oxzi (dismissed):

yhabteab:
After a short discussion with @nilmerg offline, we agreed to rename the HTTP header to `X-Icinga-Reject-If-Relations-Incomplete` to reflect its usage correctly, so I've updated the PR accordingly.

oxzi (approved):

PR #204, opened 2026-05-28
config: add TLS server config options

yhabteab:
This PR adds TLS server configuration options to the `config` package. Most of the options are extracted from the existing TLS client configuration options, and some new options are added to support server-specific features. Thus, both client and server TLS configurations are separated into their own structs, while still sharing common fields through embedding. The new server struct is named `ServerTLS` to clearly indicate its purpose and avoid confusion with the existing `TLS` struct used for client configurations.


config/tls.go
```diff
@@ -70,54 +110,123 @@ func loadPemOrFile(pemOrFile string) ([]byte, error) {
...
+	switch str {
+	case "NoClientCert":
+		*o = TlsClientAuthType(tls.NoClientCert)
+	case "RequireAnyClientCert":
+		*o = TlsClientAuthType(tls.RequireAnyClientCert)
+	case "VerifyClientCertIfGiven":
+		*o = TlsClientAuthType(tls.VerifyClientCertIfGiven)
+	case "RequireAndVerifyClientCert":
+		*o = TlsClientAuthType(tls.RequireAndVerifyClientCert)
+	default:
```
oxzi, line 157: This switch misses `tls.RequestClientCert` which is also a valid [`tls.ClientAuthType`](https://pkg.go.dev/crypto/tls#ClientAuthType). It is also not available in `ServerTLS.Validate` and I found no reference to it.

config/tls.go
```diff
@@ -70,54 +110,123 @@ func loadPemOrFile(pemOrFile string) ([]byte, error) {
...
+	ClientAuth TlsClientAuthType `yaml:"client_auth" env:"CLIENT_AUTH" default:"NoClientCert"`
+}
+
+// Validate checks the [ServerTLS] configuration for consistency and returns an error if the configuration is invalid.
+func (st *ServerTLS) Validate() error {
+	switch tls.ClientAuthType(st.ClientAuth) {
+	case tls.NoClientCert, tls.RequireAnyClientCert, tls.VerifyClientCertIfGiven, tls.RequireAndVerifyClientCert:
+		return nil
+	default:
+		return errors.Errorf("invalid ClientAuth value: %q", st.ClientAuth)
```
oxzi: Unsure if `%q` would work here as your `TlsClientAuthType` type has no `fmt.Stringer` implementation.

config/tls.go
```diff
@@ -70,54 +110,123 @@ func loadPemOrFile(pemOrFile string) ([]byte, error) {
...
+
+			tlsConfig.ClientCAs = x509.NewCertPool()
+			if !tlsConfig.ClientCAs.AppendCertsFromPEM(caPem) {
+				return nil, errors.New("can't parse CA file")
+			}
 		}
-	}
 
-	tlsConfig.ServerName = serverName
+		tlsConfig.ClientAuth = tls.ClientAuthType(st.ClientAuth)
```
oxzi, line 237: At the moment, it would be possible to set this to `RequireAndVerifyClientCert` or `VerifyClientCertIfGiven` without specifying a CA. AFAIK, this would fall back to the system's CA bundle, but this might be undesirable.

oxzi (commented):
In general, LGTM. But I would be happy about some tests, especially for the `TlsClientAuthType` parsing and validation.

config/tls.go
yhabteab: > FAIK, this would fall back to the system's CA bundle, but this might be undesirable.

Why would it be undesirable? After all, that's a config option, so it's up to the user whether he wants to explicitly provide a CA or not, and btw. that's exactly the case for clients as well (users aren't required to provide a CA even if they set `TLS.Insecure` to false.

config/tls.go
yhabteab: I actually left it out on purpose but I don't know anymore why I did that :)! I'll re-add it.

yhabteab:
> But I would be happy about some tests, especially for the `TlsClientAuthType` parsing and validation.

Done!

config/tls.go
oxzi: > Why would it be undesirable?

Because there are lots of CAs in it, and some are maybe not so trustworthy[^1]. Furthermore, if you use client certificates, you want to make sure that they are signed by your CA as otherwise the whole process would be next to useless.

However, this comment was more like a "maybe we should think about this, as this might be a security footgun" as an actual issue or task. We could consider requiring a CA if one of those options are set, or just ignore it for the moment. Dunno, sorry.

[^1]: <https://bugzilla.mozilla.org/show_bug.cgi?id=647959> for comedy.

yhabteab: Ok, changed it as discussed offline.

oxzi (approved):

PR #205, opened 2026-05-28
notifications.source: Configurable relations

oxzi:
> [!WARNING]
>
> This PR is not yet part of the main branch since it introduces a breaking change to the notifications package and all its users must be fixed first before merging this into the main branch (See the failing GHAs below).

Extend the Config struct to allow specifying default relations to always be used when submitting events.

---

This PR is required for changes in Icinga/icingadb#1120 to address the following requirement from Icinga/icingadb#1098.

> If Icinga DB is configured to always include particular relations, it can signal Icinga Notifications that these are complete and cannot be updated by using a request header: X-Icinga-Complete-Relations: host.vars, services[*].vars

### DIFF
--- a/notifications/source/config.go
+++ b/notifications/source/config.go
@@ -15,6 +15,9 @@ type Config struct {
 	// Password is the API user's password for the Icinga Notifications API.
 	Password     string `yaml:"password" env:"PASSWORD,unset"` // #nosec G117 -- exported password field
 	PasswordFile string `yaml:"password_file" env:"PASSWORD_FILE"`
+
+	// DefaultRelations to always resolve and include in the events submitted to Icinga Notifications.
+	DefaultRelations []string `yaml:"default_relations" env:"DEFAULT_RELATIONS"`
 }
 
 // Validate the configuration, implements config.Validator.
--- a/notifications/source/config_test.go
+++ b/notifications/source/config_test.go
@@ -53,6 +53,30 @@ password_file: %s`, passwordFile),
 				PasswordFile: passwordFile,
 			},
 		},
+		{
+			Name: "Custom attribute negotiation",
+			Data: testutils.ConfigTestData{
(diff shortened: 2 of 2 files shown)


notifications/source/config.go
yhabteab, line 24: `DefaultRelations` must also be JSONPath validated here.

notifications/source/config.go
```diff
@@ -15,6 +15,11 @@ type Config struct {
 	// Password is the API user's password for the Icinga Notifications API.
 	Password     string `yaml:"password" env:"PASSWORD,unset"` // #nosec G117 -- exported password field
 	PasswordFile string `yaml:"password_file" env:"PASSWORD_FILE"`
+
+	// DefaultRelations signals to always use these relations and do not let the source implementation decide.
+	DefaultRelations []string `yaml:"default_relations" env:"DEFAULT_RELATIONS"`
+	// DisableNegotiation disables the attribute negotiation between the source and Icinga Notifications.
+	DisableNegotiation bool `yaml:"disable_negotiation" env:"DISABLE_NEGOTIATION"`
```
yhabteab: Is this really required? Was it mentioned somewhere in the Icinga DB issue?

notifications/source/config.go
oxzi: Yes, but the source is going to parse this anyway.

Thus, I don't think it is necessary to pull in a JSONPath parser library as a dependency here. This will only double the amount of dependabot PRs.

notifications/source/config.go
oxzi: The issue was quite vague in this regard and used the word "can".

> it can signal Icinga Notifications that these are complete and cannot be updated

So, I decided to let the user have a feelgood switch.

oxzi: Removed.

yhabteab (approved):

PR #212, opened 2026-06-17
Miscellaneous changes for Notifications

yhabteab:
This adds the following three commits needed by the Icinga Notifications source implementation in Icinga DB and by Icinga Notifications daemon itself:

- d3bfbd9 This commit adds some new fields to the `Event` type and renames the existing `Mute` and `MuteReason` to `Muted` and `MutedReason` accordingly. Also, this commit moves some basic `Event` validations (as described in https://github.com/Icinga/icinga-notifications/issues/407) from Icinga Notifications that can be used from both the sources and Icinga Notifications to validate generated and receivent events.
- 1bdb4a5 This commit introduces a new `OnSuccessApplyAndSendTo` helper function. It is similar to the existing `OnSuccessSendTo` one, but with an additional option to apply a custom function on the received row before streaming it to out channel.
- 06c9e1b Introduces a new `Redis.XDelOnAllConsumersAck` helper function, which allows for the deletion of messages from a stream once they have been acknowledged by all consumers. This function is particularly useful in scenarios where 2 or more consumers need to xread from the same stream, and want to still ensure that messages are only deleted after all consumers have acknowledged them.

refs https://github.com/Icinga/icingadb/pull/1130


yhabteab:
Sorry, I actually forgot to drop the now longer used `Event.Type` and `Event.Username` fields.

yhabteab:
I've fixed the linter errors and rephrased some of the doc blocks a bit.

yhabteab:
Rebased.

notifications/event/event.go
```diff
@@ -72,3 +109,62 @@ type Event struct {
...
+func (e *Event) CloseIncident() bool { return e.Close.Valid && e.Close.Bool }
+
+// NotifyRecipients returns true if this event should notify the incident recipients.
+func (e *Event) NotifyRecipients() bool { return e.Notify.Valid && e.Notify.Bool }
+
+// Validate checks if the event is valid and can be processed by Icinga Notifications.
+//
+// It checks if the required fields are set and if the values of the fields are valid.
+// If any of the checks fail, it returns an error describing the issue. Otherwise, it returns nil.
+func (e *Event) Validate() error {
```
nilmerg, line 144: `{"close": true}` isn't a valid case anymore, only in combination `{"incident: true, "close": true}` it is still valid.

nilmerg (changes requested):

notifications/event/event.go
yhabteab: Fixed.

notifications/event/event.go
```diff
@@ -72,3 +109,66 @@ type Event struct {
...
```
nilmerg, line 144: And, not sure if it's correct to request the change here, but this should probably also complain about just `{"message": …}` as this wasn't a valid case to begin with.

notifications/event/event.go
```diff
@@ -72,3 +109,66 @@ type Event struct {
...
+	}
+	if !e.NotifyRecipients() && e.Notify.Valid {
+		return errors.New("invalid event: 'notify' can only be set to true or non at all, but not to false")
+	}
+
+	if !e.OpenOrEscalate() && e.CloseIncident() {
+		return errors.New("invalid event: 'close' must not be set if 'incident' is not set")
+	}
+
+	if e.IsMuted() && e.MutedReason == "" {
```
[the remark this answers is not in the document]
yhabteab: @nilmerg is this actually correct? Is `muted_reason` always required to be set if `muted` is set explicitly to either true or false?

redis/client.go
```diff
@@ -289,6 +290,74 @@ func (c *Client) XReadUntilResult(ctx context.Context, a *redis.XReadArgs) ([]re
...
+	ackCount := len(messageIDsIn)
+	if ackCount == 0 {
+		return errors.New("at least one input channel is required")
+	}
+
+	deletes := make(chan string)
+	acks := make(map[string]int)
+	var mux sync.Mutex
+
+	g, ctx := errgroup.WithContext(ctx)
```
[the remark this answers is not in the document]
yhabteab: ach, you're absolutely right..

database/db.go
```diff
@@ -428,6 +428,21 @@ func OnSuccessSendTo[T any](ch chan<- T) OnSuccess[T] {
...
+// channel of type U after applying the provided function on them.
+func OnSuccessApplyAndSendTo[T, U any](out chan<- U, fn func(T) U) OnSuccess[T] {
+	return func(ctx context.Context, affectedRows []T) error {
+		for _, affectedRow := range affectedRows {
+			select {
+			case out <- fn(affectedRow):
+			case <-ctx.Done():
+				return ctx.Err()
+			}
+		}
```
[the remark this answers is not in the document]
yhabteab: Done.

notifications/event/event.go
yhabteab: Fixed.

redis/client.go
```diff
@@ -289,6 +290,87 @@ func (c *Client) XReadUntilResult(ctx context.Context, a *redis.XReadArgs) ([]re
...
+// of input channels, and once that many acknowledgments for a message ID have been received, the message is
+// ultimately deleted from the Redis stream.
+func (c *Client) XDelOnAllConsumersAck(ctx context.Context, stream string, messageIDsIn ...<-chan string) error {
+	ackCount := len(messageIDsIn)
+	if ackCount == 0 {
+		return errors.New("at least one input channel is required")
+	}
+
+	deletes := make(chan string)
+	acks := make(map[string]int)
```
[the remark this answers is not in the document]
yhabteab: There is no such infrastructure yet, so I won't be able to add such integration tests here, so closing.

notifications/event/event.go
```diff
@@ -72,3 +109,71 @@ type Event struct {
...
+	if !e.CloseIncident() && e.Close.Valid {
+		return errors.New("invalid event: 'close' can only be set to true or none at all, but not to false")
+	}
+	if !e.NotifyRecipients() && e.Notify.Valid {
+		return errors.New("invalid event: 'notify' can only be set to true or none at all, but not to false")
+	}
+
+	if !e.OpenOrEscalate() && e.CloseIncident() {
+		return errors.New("invalid event: 'close' must not be set if 'incident' is not set")
+	}
```
oxzi, line 173: Your documentation states:

> If there's no existing incident and the Incident field isn't set, this field has no effect.

But this check aborts if Close is set and no Incident is present.

oxzi (commented):
Looks good except the one comment. My prior two comments were already addressed in the meantime :)

---

For the CI, please consider rebasing.

notifications/event/event.go
yhabteab: Right! This validation was added with the latest changes after a discussion with Johannes. I'll fix the comment.

yhabteab:
> For the CI, please consider rebasing.

For the CI? The compatibility tests are still going to fail but rebased it anyway.

oxzi:
> > For the CI, please consider rebasing.
> 
> For the CI? The compatibility tests are still going to fail but rebased it anyway.

Umm, yes, I'm still not fully awake.

---

Otherwise, the PR LGTM. Thanks.

Due to the breaking API, I am going to approve it when I am finished with the other PRs.

oxzi (approved):
🚢

PR #217, opened 2026-07-02
Support `/incidents` Icinga Notifications endpoint

yhabteab:
This PR adds support for the `/incidents` endpoint for the Notifications source client. The `/incidents` endpoint will allow users to retrieve incidents specific to a given source based on a given filter. Also, the same endpoint can be used to modify incidents of a said source. In order to support this endpoint, I had to refactor the exiting `ProcessEvent` a bit to outsource common logic to a new helper function that's now used by all of them.

refs https://github.com/Icinga/icinga-notifications/issues/452

### DIFF
--- a/notifications/source/client.go
+++ b/notifications/source/client.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/icinga/icinga-go-library/notifications"
 	"github.com/icinga/icinga-go-library/notifications/event"
+	"github.com/icinga/icinga-go-library/types"
 	"github.com/pkg/errors"
 )
 
@@ -42,8 +43,12 @@ func (b *basicAuthTransport) RoundTrip(req *http.Request) (*http.Response, error
 // It stores the configuration for the API endpoint and holds a reusable HTTP client for requests. To create a Client,
 // the NewClient function should be used.
 type Client struct {
-	httpClient           http.Client
-	processEventEndpoint string
+	httpClient http.Client
+
+	endpoints struct {
+		ProcessEvent string
+		Incidents    string
+	}
 }
 
 // NewClient creates a new Client instance with the provided configuration.
@@ -57,8 +62,8 @@ func NewClient(cfg Config, clientName string) (*Client, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "unable to parse API base URL")
 	}
-
-	processEventEndpoint := baseUrl.JoinPath("/process-event").String()
+	// Clear any query parameters from the base URL not to interfere with the filter query parameter used in requests.
+	baseUrl.RawQuery = ""
 
 	return &Client{
 		httpClient: http.Client{
@@ -69,7 +74,13 @@ func NewClient(cfg Config, clientName string) (*Client, error) {
 				userAgent:    clientName,
 			},
 		},
-		processEventEndpoint: processEventEndpoint,
+		endpoints: struct {
+			ProcessEvent string
+			Incidents    string
+		}{
+			ProcessEvent: baseUrl.JoinPath("/process-event").String(),
+			Incidents:    baseUrl.JoinPath("/incidents").String(),
+		},
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


notifications/source/client.go
```diff
@@ -78,32 +89,23 @@ var ErrAttrsNegotiation = stderrors.New("attribute negotiation required")
...
-		return nil, errors.Wrap(err, "cannot create HTTP request")
-	}
-
-	req.Header.Add("Content-Type", "application/json")
-	req.Header.Add("Accept", "application/json")
-	req.Header.Add(notifications.XIcingaRejectIfRelationsIncomplete, fmt.Sprintf("%t", rejectIncompete))
-
-	resp, err := client.httpClient.Do(req) // #nosec G704 -- SSRF impossible, trusted user input
+	//nolint:bodyclose // False positive, drainBody is called in the defer statement below.
+	resp, err := client.doRequest(ctx, http.MethodPost, client.endpoints.ProcessEvent, nil, bytes.NewReader(body), map[string]string{
```
oxzi: Please add a few newlines here to keep the line width below 120 characters and to ease the readability. Same applies to a few other lines below, such as 181, 205, and 259.

notifications/source/client.go
```diff
@@ -123,9 +125,148 @@ func (client *Client) ProcessEvent(ctx context.Context, ev *event.Event, rejectI
 		return nil, nil // Superfluous state change event.
 	}
 
-	var buf bytes.Buffer
-	_, _ = io.Copy(&buf, &io.LimitedReader{R: resp.Body, N: 1 << 16}) // Limit the error message length to avoid memory exhaustion.
-
 	return nil, errors.Errorf("unexpected response from process event API, status %q (%d): %q",
```
oxzi, line 136: This has nothing to do with this PR, but could you check for the 401 response code and add a custom exported error type for it while editing this file? This would enable us to detect it in other places, such as Icinga/icingadb#1142. This should be done for each endpoint. Thanks!

notifications/source/client.go
```diff
@@ -123,9 +125,148 @@ func (client *Client) ProcessEvent(ctx context.Context, ev *event.Event, rejectI
...
+}
+
+// GetIncidents retrieves the current incidents of this source from the Icinga Notifications /incidents API endpoint.
+//
+// The filter parameter is used to filter the incidents returned by the API. It must be a non-nil value that can
+// be marshaled to JSON. If filter is nil, an error is returned. Please refer to the Icinga Notifications API doc
+// for details on the filter syntax and supported fields.
+func (client *Client) GetIncidents(ctx context.Context, filter any) ([]Incident, error) {
+	if filter == nil {
+		return nil, errors.New("you must provide a non-nil filter")
```
oxzi: I would expect an error message to be in the passive voice, or at least not address me directly. In the end, this will show up in the logs, and some user will feel confronted even though they did not do anything themselves.

notifications/source/client.go
```diff
@@ -123,9 +125,148 @@ func (client *Client) ProcessEvent(ctx context.Context, ev *event.Event, rejectI
...
+}
+
+// ModifiableIncidentAttrs represents the attributes of an incident that can be modified via the /incident endpoint.
+//
+// The Message field is a string that can be set to update the incident's message. The Close field is a boolean
+// that can only be set to true to close the incident. Setting Close to false is not allowed and will result in
+// a validation error.
+type ModifiableIncidentAttrs struct {
+	Message types.String `json:"message"`
+	Close   types.Bool   `json:"close"`
```
oxzi:
```suggestion
	Message types.String `json:"message,omitzero"`
	Close   types.Bool   `json:"close,omitzero"`
```

Otherwise, a `null` field will be sent as well. Look at the following test: https://go.dev/play/p/GhuURa30OuV

As the issue wrote about "modifications [being] allowed", I would expect them to be absent if unset.

notifications/source/client.go
```diff
@@ -123,9 +125,148 @@ func (client *Client) ProcessEvent(ctx context.Context, ev *event.Event, rejectI
...
+}
+
+// readLimitedBody reads from the provided [io.ReadCloser] up to 1<<16 bytes and returns it as a string.
+//
+// If an error occurs during reading, the error message is appended to the returned string.
+func readLimitedBody(body io.ReadCloser) string {
+	buf := new(bytes.Buffer)
+	_, err := io.Copy(buf, &io.LimitedReader{R: body, N: 1 << 16}) // Limit the error message length to avoid memory exhaustion.
+	if err != nil {
+		buf.WriteString(" Read Error: " + err.Error())
```
oxzi: Nitpick forwarded from my editor: "Inefficient string concatenation in call to WriteString".

Which makes sense, as the `+` creates a string on the heap just to be passed on. Thus, maybe make it two `buf.WriteString` calls.

notifications/source/client.go
```diff
@@ -78,32 +89,23 @@ var ErrAttrsNegotiation = stderrors.New("attribute negotiation required")
...
 		return nil, errors.Wrap(err, "cannot encode event to JSON")
 	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, client.processEventEndpoint, bytes.NewReader(body))
-	if err != nil {
-		return nil, errors.Wrap(err, "cannot create HTTP request")
-	}
-
-	req.Header.Add("Content-Type", "application/json")
-	req.Header.Add("Accept", "application/json")
```
oxzi, line 95: The `Accept` header is no longer being sent. Unsure if we need it.

yhabteab: I actually removed it on purpose because I was thinking that it shouldn't be needed for anything, but I'll re-add it again since this PR is not related to taht. If in doubt it should be removed in a separate PR or we just leave it as-is.

notifications/source/client.go
yhabteab: Reformatted the lines.

notifications/source/client.go
yhabteab: Done.

notifications/source/client.go
yhabteab: Rephrased it.

notifications/source/client.go
yhabteab: Done.

notifications/source/client.go
yhabteab: Done.

notifications/source/client.go
oxzi: Merci

oxzi (approved):

PR #219, opened 2026-07-08
add CRL-based client certificate revocation check

### LINKED ISSUE #218: Add client certificate revocation checking
Currently, icinga-go-library has no support for CRL-based client certificate revocation. This should be implemented centrally so all Icinga components can use it without duplicating the logic.

Related: 
- https://github.com/Icinga/icinga-notifications/issues/411

Jan-Schuppik:
Add a crlChecker that loads a PEM or DER Certificate Revocation List, verifies its signature against the CA, and watches the file for live updates via fsnotify. Atomic file replacements (mv) are handled by re-adding the watch path on Create/Rename events.

Expose ApplyRevocationCheck on TLSCommon, which wires the checker into tls.Config.VerifyConnection so revoked client certificates are rejected at handshake time.

resolves #218

refs https://github.com/Icinga/icinga-notifications/pull/454

### DIFF
--- a/config/crl.go
+++ b/config/crl.go
@@ -1,0 +1,128 @@
+package config
+
+import (
+	"context"
+	"crypto/x509"
+	"encoding/pem"
+	"fmt"
+	"math/big"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	"go.uber.org/zap"
+)
+
+// CrlChecker loads a Certificate Revocation List (CRL) from a file, verifies its signature
+// against a trusted CA certificate, and exposes [CrlChecker.IsRevoked] for serial number
+// lookups. The in-memory CRL is protected by a read/write mutex so that background reloads
+// via [CrlChecker.WatchAndReload] are safe to run concurrently with ongoing checks.
+type CrlChecker struct {
+	mu         sync.RWMutex
+	path       string
+	issuer     *x509.Certificate
+	crlExpiry  time.Time
+	crlRevoked map[string]struct{}
+}
+
+// NewCRLChecker creates a [CrlChecker] by loading the CRL at path and verifying its signature
+// against issuer. The file may be PEM- or DER-encoded. An error is returned if the file cannot
+// be read, is not a valid CRL, or its signature does not match issuer.
+func NewCRLChecker(path string, issuer *x509.Certificate) (*CrlChecker, error) {
+	c := &CrlChecker{path: path, issuer: issuer}
+	if err := c.reload(); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+// reload reads the CRL file, verifies its signature against the issuer, and updates the in-memory CRL.
+func (c *CrlChecker) reload() error {
+	data, err := os.ReadFile(c.path)
+	if err != nil {
+		return fmt.Errorf("cannot read CRL file: %w", err)
+	}
+	if block, _ := pem.Decode(data); block != nil {
+		data = block.Bytes // strip PEM wrapper if present; ParseRevocationList wants DER
+	}
+	rl, err := x509.ParseRevocationList(data)
+	if err != nil {
+		return fmt.Errorf("cannot parse CRL: %w", err)
+	}
+
+	err = rl.CheckSignatureFrom(c.issuer)
+	if err != nil {
+		return fmt.Errorf("CRL signature invalid: %w", err)
+	}
+
+	revoked := make(map[string]struct{})
+	for _, entry := range rl.RevokedCertificateEntries {
+		revoked[entry.SerialNumber.Text(16)] = struct{}{}
+	}
+
+	c.mu.Lock()
+	c.crlExpiry = rl.NextUpdate
+	c.crlRevoked = revoked
+	c.mu.Unlock()
+
+	return nil
+}
+
+// WatchAndReload blocks, watching the CRL file for changes and reloading it whenever a
+// Write, Create, or Rename event is received. It returns when ctx is canceled.
+// Reload failures are logged as warnings and do not stop the watcher.
+func (c *CrlChecker) WatchAndReload(ctx context.Context, logger *zap.SugaredLogger) error {
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return fmt.Errorf("cannot create watcher: %w", err)
+	}
+	defer func() { _ = watcher.Close() }()
+	if err := watcher.Add(filepath.Dir(c.path)); err != nil {
+		return fmt.Errorf("cannot watch CRL file: %w", err)
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case event, ok := <-watcher.Events:
+			if !ok {
+				return fmt.Errorf("CRL watcher channel closed unexpectedly")
+			}
+
+			if filepath.Base(event.Name) != filepath.Base(c.path) {
+				continue
+			}
+
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) || event.Has(fsnotify.Rename) {
+				if err := c.reload(); err != nil {
+					logger.Warnw("CRL reload failed", zap.Error(err))
+				} else {
+					logger.Info("CRL reloaded successfully")
+				}
+			}
+		case err, ok := <-watcher.Errors:
+			if !ok {
+				return fmt.Errorf("CRL watcher error channel closed unexpectedly")
+			}
+			logger.Warnw("CRL watcher error", zap.Error(err))
+		}
+	}
+}
+
+// IsRevoked reports whether serial appears in the CRL.
+//
+// It returns an error if the CRL is outdated (NextUpdate in the past).
+func (c *CrlChecker) IsRevoked(serial *big.Int) (bool, error) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if time.Now().After(c.crlExpiry) {
+		return false, fmt.Errorf("CRL is outdated (NextUpdate=%s)", c.crlExpiry)
+	}
+
+	_, revoked := c.crlRevoked[serial.Text(16)]
+	return revoked, nil
+}
--- a/config/crl_test.go
+++ b/config/crl_test.go
@@ -1,0 +1,291 @@
+package config
+
+import (
+	"bytes"
+	"context"
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"crypto/x509"
+	"crypto/x509/pkix"
+	"encoding/pem"
+	"math/big"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"go.uber.org/zap/zaptest"
+)
+
+func testCA(t *testing.T) (*x509.Certificate, crypto.PrivateKey) {
+	key, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	require.NoError(t, err)
+	serial, err := rand.Int(rand.Reader, new(big.Int).Lsh(big.NewInt(1), 128))
+	require.NoError(t, err)
+	template := &x509.Certificate{
+		SerialNumber:          serial,
+		Subject:               pkix.Name{CommonName: "test-ca"},
+		NotBefore:             time.Now().Add(-time.Hour),
+		NotAfter:              time.Now().Add(24 * time.Hour),
+		KeyUsage:              x509.KeyUsageCertSign | x509.KeyUsageCRLSign,
+		BasicConstraintsValid: true,
+		IsCA:                  true,
+	}
+	der, err := x509.CreateCertificate(rand.Reader, template, template, key.Public(), key)
+	require.NoError(t, err)
+	cert, err := x509.ParseCertificate(der)
+	require.NoError(t, err)
+	return cert, key
+}
+
+// buildCRL builds a CRL with the given parameters and returns its PEM-encoded bytes.
+func buildCRL(t *testing.T, ca *x509.Certificate, caKey crypto.PrivateKey, nextUpdate time.Time, revokedSerials ...*big.Int) []byte {
+	entries := make([]x509.RevocationListEntry, 0, len(revokedSerials))
+	for _, s := range revokedSerials {
+		entries = append(entries, x509.RevocationListEntry{
+			SerialNumber:   s,
+			RevocationTime: time.Now().Add(-time.Hour),
+		})
+	}
+	template := &x509.RevocationList{
+		Number:                    big.NewInt(1),
+		ThisUpdate:                time.Now().Add(-time.Hour),
+		NextUpdate:                nextUpdate,
+		RevokedCertificateEntries: entries,
+	}
+	signer, ok := caKey.(crypto.Signer)
+	require.True(t, ok, "caKey must implement crypto.Signer")
+	der, err := x509.CreateRevocationList(rand.Reader, template, ca, signer)
+	require.NoError(t, err)
+	var buf bytes.Buffer
+	require.NoError(t, pem.Encode(&buf, &pem.Block{Type: "X509 CRL", Bytes: der}))
+	return buf.Bytes()
+}
+
+// createCRLFile creates a temporary CRL file and returns its path.
+func createCRLFile(t *testing.T, ca *x509.Certificate, caKey crypto.PrivateKey, nextUpdate time.Time, revokedSerials ...*big.Int) string {
+	tmp, err := os.CreateTemp(t.TempDir(), "crl-*.pem")
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = os.Remove(tmp.Name()) })
+	_, err = tmp.Write(buildCRL(t, ca, caKey, nextUpdate, revokedSerials...))
+	require.NoError(t, err)
+	require.NoError(t, tmp.Close())
+	return tmp.Name()
+}
+
+// atomicReplaceCRL writes a new CRL to a temp file then renames it over path,
+// mirroring the production mv-based CRL update that exercises the fsnotify re-add branch.
+func atomicReplaceCRL(t *testing.T, path string, ca *x509.Certificate, caKey crypto.PrivateKey, nextUpdate time.Time, revokedSerials ...*big.Int) {
+	_ = os.Remove(path)
+	tmp, err := os.CreateTemp(filepath.Dir(path), "crl-new-*.pem")
+	require.NoError(t, err)
+	_, err = tmp.Write(buildCRL(t, ca, caKey, nextUpdate, revokedSerials...))
+	require.NoError(t, err)
+	require.NoError(t, tmp.Close())
+	require.NoError(t, os.Rename(tmp.Name(), path))
+
+	time.Sleep(200 * time.Millisecond)
+}
+
+func TestNewCRLChecker(t *testing.T) {
+	ca, caKey := testCA(t)
+	future := time.Now().Add(24 * time.Hour)
+
+	t.Run("Valid PEM CRL", func(t *testing.T) {
+		path := createCRLFile(t, ca, caKey, future)
+		checker, err := NewCRLChecker(path, ca)
+		require.NoError(t, err)
+		require.NotNil(t, checker)
+	})
+
+	t.Run("Valid DER CRL", func(t *testing.T) {
+		template := &x509.RevocationList{
+			Number:     big.NewInt(1),
+			ThisUpdate: time.Now().Add(-time.Hour),
+			NextUpdate: future,
+		}
+		signer, ok := caKey.(crypto.Signer)
+		require.True(t, ok, "caKey must implement crypto.Signer")
+		der, err := x509.CreateRevocationList(rand.Reader, template, ca, signer)
+		require.NoError(t, err)
+		f, err := os.CreateTemp(t.TempDir(), "crl-*.der")
+		require.NoError(t, err)
+		t.Cleanup(func() { _ = os.Remove(f.Name()) })
+		_, err = f.Write(der)
+		require.NoError(t, err)
+		require.NoError(t, f.Close())
+
+		checker, err := NewCRLChecker(f.Name(), ca)
+		require.NoError(t, err)
+		require.NotNil(t, checker)
+	})
+
+	t.Run("Nonexistent file", func(t *testing.T) {
+		_, err := NewCRLChecker("/nonexistent/crl.pem", ca)
+		require.ErrorContains(t, err, "cannot read CRL file")
+	})
+
+	t.Run("Corrupt data", func(t *testing.T) {
+		f, err := os.CreateTemp(t.TempDir(), "crl-*.pem")
+		require.NoError(t, err)
+		t.Cleanup(func() { _ = os.Remove(f.Name()) })
+		_, err = f.Write([]byte("not a crl"))
+		require.NoError(t, err)
+		require.NoError(t, f.Close())
+
+		_, err = NewCRLChecker(f.Name(), ca)
+		require.ErrorContains(t, err, "cannot parse CRL")
+	})
+
+	t.Run("Wrong CA signature", func(t *testing.T) {
+		otherCA, otherKey := testCA(t)
+		path := createCRLFile(t, otherCA, otherKey, future)
+		_, err := NewCRLChecker(path, ca)
+		require.ErrorContains(t, err, "CRL signature invalid")
+	})
+}
+
+func TestCRLChecker_IsRevoked(t *testing.T) {
+	ca, caKey := testCA(t)
+	revokedSerial := big.NewInt(42)
+	validSerial := big.NewInt(43)
+	future := time.Now().Add(24 * time.Hour)
+
+	path := createCRLFile(t, ca, caKey, future, revokedSerial)
+	checker, err := NewCRLChecker(path, ca)
+	require.NoError(t, err)
+
+	t.Run("Revoked serial", func(t *testing.T) {
+		revoked, err := checker.IsRevoked(revokedSerial)
+		require.NoError(t, err)
+		require.True(t, revoked)
+	})
+
+	t.Run("Non-revoked serial", func(t *testing.T) {
+		revoked, err := checker.IsRevoked(validSerial)
+		require.NoError(t, err)
+		require.False(t, revoked)
+	})
+
+	t.Run("Expired CRL", func(t *testing.T) {
+		// Must be after ThisUpdate (-1h) but before now so the expiry branch triggers.
+		past := time.Now().Add(-30 * time.Minute)
+
+		// Build checker with expired CRL that contains revokedSerial.
+		expiredPath := createCRLFile(t, ca, caKey, past, revokedSerial)
+		expiredChecker, err := NewCRLChecker(expiredPath, ca)
+		require.NoError(t, err)
+
+		// Replace with an expired CRL that no longer lists revokedSerial.
+		atomicReplaceCRL(t, expiredPath, ca, caKey, past)
+
+		// IsRevoked detects expiry, throws an error even though the serial is in the CRL.
+		_, err = expiredChecker.IsRevoked(revokedSerial)
+		require.ErrorContains(t, err, "CRL is outdated (NextUpdate=")
+	})
+}
+
+func TestCRLChecker_WatchAndReload(t *testing.T) {
+	ca, caKey := testCA(t)
+	revokedSerial := big.NewInt(42)
+	future := time.Now().Add(24 * time.Hour)
+	logger := zaptest.NewLogger(t).Sugar()
+
+	t.Run("Reloads on atomic file replacement", func(t *testing.T) {
+		path := createCRLFile(t, ca, caKey, future, revokedSerial)
+		checker, err := NewCRLChecker(path, ca)
+		require.NoError(t, err)
+
+		revoked, err := checker.IsRevoked(revokedSerial)
+		require.NoError(t, err)
+		require.True(t, revoked)
+
+		ctx, cancel := context.WithCancel(t.Context())
+		defer cancel()
+
+		go func() {
+			if err := checker.WatchAndReload(ctx, logger); err != nil {
+				require.ErrorIs(t, err, context.Canceled)
+			}
+		}()
+		time.Sleep(1 * time.Second)
+
+		// Simulate production CRL rotation via atomic rename.
+		atomicReplaceCRL(t, path, ca, caKey, future)
+
+		revoked, err = checker.IsRevoked(revokedSerial)
+		require.NoError(t, err)
+		require.False(t, revoked)
+
+		atomicReplaceCRL(t, path, ca, caKey, future, revokedSerial)
+
+		revoked, err = checker.IsRevoked(revokedSerial)
+		require.NoError(t, err)
+		require.True(t, revoked)
+	})
+
+	t.Run("Deleted file fallback", func(t *testing.T) {
+		path := createCRLFile(t, ca, caKey, future, revokedSerial)
+		checker, err := NewCRLChecker(path, ca)
+		require.NoError(t, err)
+
+		revoked, err := checker.IsRevoked(revokedSerial)
+		require.NoError(t, err)
+		require.True(t, revoked)
+
+		ctx, cancel := context.WithCancel(t.Context())
+		defer cancel()
+
+		go func() {
+			if err := checker.WatchAndReload(ctx, logger); err != nil {
+				require.ErrorIs(t, err, context.Canceled)
+			}
+		}()
+
+		time.Sleep(100 * time.Millisecond)
+
+		_ = os.Remove(path)
+
+		revoked, err = checker.IsRevoked(revokedSerial)
+		require.NoError(t, err)
+		require.True(t, revoked)
+	})
+
+	t.Run("Cached CRL file expires", func(t *testing.T) {
+		path := createCRLFile(t, ca, caKey, time.Now().Add(2*time.Second), revokedSerial)
+		checker, err := NewCRLChecker(path, ca)
+		require.NoError(t, err)
+
+		ctx, cancel := context.WithCancel(t.Context())
+		defer cancel()
+
+		go func() {
+			if err := checker.WatchAndReload(ctx, logger); err != nil {
+				require.ErrorIs(t, err, context.Canceled)
+			}
+		}()
+		time.Sleep(100 * time.Millisecond)
+
+		revoked, err := checker.IsRevoked(revokedSerial)
+		require.NoError(t, err)
+		require.True(t, revoked)
+
+		_ = os.Remove(path)
+
+		revoked, err = checker.IsRevoked(revokedSerial)
+		require.NoError(t, err)
+		require.True(t, revoked)
+
+		require.Eventually(t, func() bool {
+			_, err := checker.IsRevoked(revokedSerial)
+			if err == nil {
+				return false
+			}
+			return strings.Contains(err.Error(), "CRL is outdated")
+		}, 5*time.Second, 100*time.Millisecond)
+	})
+}
--- a/config/tls.go
+++ b/config/tls.go
@@ -4,8 +4,10 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"encoding/pem"
-	"github.com/pkg/errors"
+	"fmt"
 	"os"
+
+	"github.com/pkg/errors"
 )
 
 // TLSCommon represents common TLS config options that are shared between TLS client and server settings.
@@ -188,6 +190,20 @@ type ServerTLS struct {
 	// Valid values are all the [tls.ClientAuthType] options typed out as strings:
 	// "NoClientCert", "RequestClientCert", "RequireAnyClientCert", "VerifyClientCertIfGiven", and "RequireAndVerifyClientCert".
 	ClientAuth TlsClientAuthType `yaml:"client_auth" env:"CLIENT_AUTH" default:"NoClientCert"`
+
+	// CrlFile is the path to the Certificate Revocation List (CRL) file.
+	//
+	// If specified, the CRL is used to check for revoked certificates in TLS.
+	// The CRL must be signed by the CA specified in the Ca option. If the CRL file is not found or cannot be loaded,
+	// an error is returned during TLS configuration.
+	// This option is ignored if TLS is not enabled. If the Ca option is not set it returns an error.
+	//
+	// At the moment, only one CA certificate is supported.
+	//
+	// Note that populating this field alone does not enable CRL checking.
+	// [ServerTLS.InitRevocationChecking] must be called after assembling the [tls.Config]
+	// to wire up the actual revocation check.
+	CrlFile string `yaml:"crl_file" env:"CRL_FILE"`
 }
 
 // Validate checks the [ServerTLS] configuration for consistency and returns an error if the configuration is invalid.
@@ -198,6 +214,9 @@ func (st *ServerTLS) Validate() error {
 
 	switch cat := tls.ClientAuthType(st.ClientAuth); cat {
 	case tls.NoClientCert, tls.RequestClientCert, tls.RequireAnyClientCert:
+		if st.CrlFile != "" {
+			return errors.New("CRL file given, but given ClientAuth mode doesn't verify client certificates")
+		}
 		// These ClientAuth types do not require a CA certificate to be configured,
 		// since the server will not verify client certificates in these modes.
 		return nil
@@ -216,6 +235,9 @@ func (st *ServerTLS) Validate() error {
 // It returns a configured *tls.Config or an error if there are issues with the provided TLS settings.
 // If TLS is not enabled (st.Enable is false), it returns nil without an error.
 func (st *ServerTLS) MakeConfig() (*tls.Config, error) {
+	if st.CrlFile != "" && st.Ca == "" {
+		return nil, errors.New("CRL file given, but CA certificate missing")
+	}
 	tlsConfig, err := st.makeConfig()
 	if err != nil {
 		return nil, err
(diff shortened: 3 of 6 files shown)


config/tls.go
```diff
@@ -35,6 +38,9 @@ type TLSCommon struct {
 	// or to verify client certificates in TLS server mode. Otherwise, the system's root CA pool is used.
 	// This option is ignored if Insecure is true in TLS client mode.
 	Ca string `yaml:"ca" env:"CA"`
+
+	// CrlFile is the path to the Certificate Revocation List (CRL) file.
+	CrlFile string `yaml:"crl_file" env:"CRL_FILE"`
```
yhabteab: This should also allow users to provide raw PM-encoded CRL string via an environment vairable and shouldn't require it to always be a file just like the other existing options. The field should be renamed to just `Crl` then. The fswatcher also need to be started conditionally, i.e., only when a CRL file is provided.

config/tls_test.go
```diff
@@ -416,6 +417,111 @@ func TestTlsClientAuthType(t *testing.T) {
...
+
+	revokedCert, _, err := generateCert("revoked", generateCertOptions{issuer: ca, issuerKey: caKey})
+	require.NoError(t, err)
+	validCert, _, err := generateCert("valid", generateCertOptions{issuer: ca, issuerKey: caKey})
+	require.NoError(t, err)
+
+	emptyCrlFile := generateCRL(t, ca, caKey, nil)
+	revokedCrlFile := generateCRL(t, ca, caKey, []*big.Int{revokedCert.SerialNumber})
+
+	t.Run("Nil tlsConfig returns nil", func(t *testing.T) {
```
yhabteab:
```suggestion
	t.Run("Nil tlsConfig", func(t *testing.T) {
```

Same with the two test cases below.

config/tls_test.go
```diff
@@ -416,6 +417,111 @@ func TestTlsClientAuthType(t *testing.T) {
...
+		require.Error(t, err)
+	})
+
+	t.Run("Nonexistent CRL file", func(t *testing.T) {
+		tc := &TLSCommon{Ca: caStr, CrlFile: "/nonexistent/crl.pem"}
+		_, err := tc.InitRevocationChecking(&tls.Config{})
+		require.ErrorContains(t, err, "cannot load CRL")
+	})
+
+	t.Run("Returns CrlChecker and sets VerifyConnection", func(t *testing.T) {
```
yhabteab:
```suggestion
	t.Run("Valid CA and Valid CRL", func(t *testing.T) {
```

config/tls_test.go
```diff
@@ -416,6 +417,111 @@ func TestTlsClientAuthType(t *testing.T) {
...
+		tc := &TLSCommon{Ca: caStr, CrlFile: emptyCrlFile}
+		tlsConf := &tls.Config{}
+		checker, err := tc.InitRevocationChecking(tlsConf)
+		require.NoError(t, err)
+		require.NotNil(t, checker)
+		require.NotNil(t, tlsConf.VerifyConnection)
+	})
+
+	t.Run("VerifyConnection", func(t *testing.T) {
+		t.Run("Empty VerifiedChains skips check", func(t *testing.T) {
```
yhabteab:
```suggestion
		t.Run("Empty VerifiedChains", func(t *testing.T) {
```

config/tls_test.go
```diff
@@ -416,6 +417,111 @@ func TestTlsClientAuthType(t *testing.T) {
...
+			tc := &TLSCommon{Ca: caStr, CrlFile: emptyCrlFile}
+			tlsConf := &tls.Config{}
+			_, err := tc.InitRevocationChecking(tlsConf)
+			require.NoError(t, err)
+			require.NoError(t, tlsConf.VerifyConnection(tls.ConnectionState{}))
+		})
+
+		t.Run("Valid cert is not rejected", func(t *testing.T) {
```
yhabteab:
```suggestion
		t.Run("Valid Cert", func(t *testing.T) {
```

config/tls_test.go
```diff
@@ -416,6 +417,111 @@ func TestTlsClientAuthType(t *testing.T) {
...
+			tlsConf := &tls.Config{}
+			_, err := tc.InitRevocationChecking(tlsConf)
+			require.NoError(t, err)
+			err = tlsConf.VerifyConnection(tls.ConnectionState{
+				VerifiedChains: [][]*x509.Certificate{{validCert}},
+			})
+			require.NoError(t, err)
+		})
+
+		t.Run("Revoked cert is rejected", func(t *testing.T) {
```
yhabteab:
```suggestion
		t.Run("Revoked Cert", func(t *testing.T) {
```

config/tls_test.go
```diff
@@ -468,3 +579,39 @@ func generateCert(cn string, options generateCertOptions) (*x509.Certificate, cr
 
 	return cert, privateKey, nil
 }
+
+func generateCRL(t *testing.T, ca *x509.Certificate, caKey crypto.PrivateKey, revokedSerials []*big.Int) string {
```
yhabteab:
```suggestion
func generateCRL(t *testing.T, ca *x509.Certificate, caKey crypto.PrivateKey, revokedSerials ...*big.Int) string {
```

That way you don't have to pass `nil` as an argument if you don't have any revokedSerials and if you do have one, then you can just pass it as-is without having to convert it to a slice.

config/tls_test.go
```diff
@@ -468,3 +579,39 @@ func generateCert(cn string, options generateCertOptions) (*x509.Certificate, cr
...
+	t.Helper()
```
yhabteab: You know what to do with this :). Same goes for the `crl_test.go` file as well.

utils/crl.go
```diff
@@ -0,0 +1,145 @@
...
+		return fmt.Errorf("cannot create watcher: %w", err)
+	}
+	if err := watcher.Add(c.path); err != nil {
+		if err := watcher.Close(); err != nil {
+			return fmt.Errorf("cannot close watcher: %w", err)
+		}
+		return fmt.Errorf("cannot watch CRL file: %w", err)
+	}
+
+	go func() {
```
yhabteab: I would not start the goroutine like this here but let the caller start it however he wants. For example, some callers may want to it to run an existing errorgroup or the like, thus I would just remove this extra goroutine here. The Icinga Notifications listener can then start it like `go crlChecker.WatchAndReload(ctx, l.logger.SugaredLogger)`.

utils/crl.go
```diff
@@ -0,0 +1,145 @@
...
+		defer func() {
+			if err := watcher.Close(); err != nil {
+				logger.Warn("cannot close watcher", zap.Error(err))
+			}
```
yhabteab: I wouldn't care much about the closing error here, so this should be just fine. Same goes for the other `watcher.Close()` call above.

```suggestion
		defer func() { _ = watcher.Close() }()
```

config/tls.go
Jan-Schuppik: As discussed offline, it doesn't make much sense to offer the option to pass the CRL as static content, since it behaves dynamically.

config/tls_test.go
yhabteab: > Same with the two test cases below.

Those two tests are still not changed.

utils/crl.go
```diff
@@ -0,0 +1,137 @@
...
+			if !ok {
+				return fmt.Errorf("CRL watcher channel closed unexpectedly")
+			}
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) || event.Has(fsnotify.Rename) || event.Has(fsnotify.Remove) {
+				// Re-add path on every event: on Linux (inotify), atomic file
+				// replacement (mv tmp.crl ca.crl) fires IN_DELETE_SELF (Remove) on
+				// the old inode, not Rename — the watcher must re-add the path to
+				// pick up the new inode. On macOS (kqueue) the same operation fires
+				// Rename instead. Handling all four events keeps both platforms working.
+				_ = watcher.Add(c.path)
```
yhabteab: Since I didn't understand what this comment is all about at first, I've exprimented a bit [in my own branch](https://github.com/Icinga/icinga-go-library/compare/add-client-cert-revoke-check...yhabteab/crl) but still don't get what the `fsnotify.Create` event is needed for. If the configured file doesn't exist then you shouldn't even be able to add it to the watch list and would already crash on the first `watcher.Add()` call on line 78, so I don't know what this event is good for. Please see the changes I made in the other branch and verify if that should cover all use cases we need.

yhabteab (commented):
I've nothing else to complain about here apart from the below inline comment but since this was requested by Alvar and Johannes, I'll let them review and approve it.

/cc @oxzi

config/tls.go
```diff
@@ -35,6 +38,14 @@ type TLSCommon struct {
...
 	// This option is ignored if Insecure is true in TLS client mode.
 	Ca string `yaml:"ca" env:"CA"`
+
+	// CrlFile is the path to the Certificate Revocation List (CRL) file.
+	//
+	// If specified, the CRL is used to check for revoked certificates in TLS.
+	// The CRL must be signed by the CA specified in the Ca option. If the CRL file is not found or cannot be loaded,
+	// an error is returned during TLS configuration.
+	// This option is ignored if TLS is not enabled or if the Ca option is not set.
+	CrlFile string `yaml:"crl_file" env:"CRL_FILE"`
```
oxzi: Why is this in `TLSCommon` and not in `ServerTLS`?

Furthermore, there should be a check in the `Validate` method to ensure that iff `CrlFile` is set, `Ca` must be set as well. Also, it must be ensured that either `tls.VerifyClientCertIfGiven` or `tls.RequireAndVerifyClientCert` are the `tls.ClientAuthType`, as otherwise this code would be a noop - but without any visible failure to the client.

config/tls.go
```diff
@@ -79,6 +90,67 @@ func (tc *TLSCommon) makeConfig() (*tls.Config, error) {
...
+	existing := tlsConfig.VerifyConnection
+	tlsConfig.VerifyConnection = func(cs tls.ConnectionState) error {
+		if existing != nil {
+			if err := existing(cs); err != nil {
+				return err
+			}
+		}
+
+		if len(cs.VerifiedChains) == 0 {
+			return nil // no client cert presented — skip
```
oxzi:
```suggestion
			return nil // no client cert presented - skip
```

utils/crl.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return false, nil
+}
+
+// getCRL returns the current CRL under a read lock.
+// It is used internally to ensure that the CRL is not modified while being accessed.
+func (c *CrlChecker) getCRL() *x509.RevocationList {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.crl
```
oxzi: This read lock only ensures that there are no races between acquiring the `crl` and setting it in the `reload` method. However, the caller might proceed with an outdated CRL.

As this method is only used in `IsRevoked`, why not remove it, and create a read lock in `IsRevoked` while iterating over the entries?

config/tls.go
```diff
@@ -79,6 +90,67 @@ func (tc *TLSCommon) makeConfig() (*tls.Config, error) {
...
+	if tlsConfig == nil || tc.CrlFile == "" || tc.Ca == "" {
+		return nil, nil
+	}
+
+	caPem, err := loadPemOrFile(tc.Ca)
+	if err != nil {
+		return nil, err
+	}
+
+	block, _ := pem.Decode(caPem)
```
oxzi: According to `Decode`'s API doc, `block` might be `nil`. Passing `nil` to `x509.ParseCertificate` might panic. Please add a `nil` check against invalid CA certificates. Btw, this check is present in `CrlChecker.reload`.

utils/crl.go
```diff
@@ -0,0 +1,137 @@
...
+			if !ok {
+				return fmt.Errorf("CRL watcher error channel closed unexpectedly")
+			}
+			logger.Warnw("CRL watcher error", zap.Error(err))
+		}
+	}
+}
+
+// IsRevoked reports whether serial appears in the CRL. If the CRL's NextUpdate time has
+// passed, the file is reloaded before the lookup so the check always operates on fresh data.
```
oxzi: > If the CRL's NextUpdate time has passed, the file is reloaded before the lookup so the check always operates on fresh data.

This does not happen. Your code returns an error.

utils/crl.go
```diff
@@ -0,0 +1,137 @@
...
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return fmt.Errorf("cannot create watcher: %w", err)
+	}
+	if err := watcher.Add(c.path); err != nil {
+		_ = watcher.Close()
+		return fmt.Errorf("cannot watch CRL file: %w", err)
+	}
+
+	defer func() { _ = watcher.Close() }()
```
oxzi:
```suggestion
	if err != nil {
		return fmt.Errorf("cannot create watcher: %w", err)
	}
	defer func() { _ = watcher.Close() }()
	
	if err := watcher.Add(c.path); err != nil {
		return fmt.Errorf("cannot watch CRL file: %w", err)
	}
```

config/crl.go
```diff
@@ -0,0 +1,137 @@
...
+				return fmt.Errorf("CRL watcher channel closed unexpectedly")
+			}
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) || event.Has(fsnotify.Rename) || event.Has(fsnotify.Remove) {
+				// Re-add path on every event: on Linux (inotify), atomic file
+				// replacement (mv tmp.crl ca.crl) fires IN_DELETE_SELF (Remove) on
+				// the old inode, not Rename — the watcher must re-add the path to
+				// pick up the new inode. On macOS (kqueue) the same operation fires
+				// Rename instead. Handling all four events keeps both platforms working.
+				_ = watcher.Add(c.path)
+				if err := c.reload(); err != nil {
```
oxzi, line 100: At least for a `fsnotify.Remove` event, the `reload` method is likely to fail - unless there is a race due to multiple similar events.

utils/crl.go
oxzi: I guess the `Create` event might be relevant in the following scenario:

1. Status quo: A CRL lies on the disk.
2. The sysadmin `scp`s a new CRL to the machine, stored somewhere in its home directory or in the temp.
3. The sysadmin removes the old CRL - the `Remove` event triggers
4. The sysadmin moves the CRL to the location - the `Create` should trigger again

However, this requires the changes from https://github.com/Icinga/icinga-go-library/pull/219/changes#r3569471676 to work.

oxzi: For a `fsnotify.Remove` event, this will error out. Afterwards there is nothing to be watched left!

Take the following demo as an example.

```go
package main

import (
        "fmt"

        "github.com/fsnotify/fsnotify"
)

func errCheck(err error) {
        if err != nil {
                panic(err)
        }
}

func main() {
        const f = "test"

        watcher, err := fsnotify.NewWatcher()
        errCheck(err)
        defer func() { _ = watcher.Close() }()

        errCheck(watcher.Add(f))
        fmt.Printf("Starting wiht: %v\n", watcher.WatchList())

        for event := range watcher.Events {
                fmt.Printf("%+v\nAdd: %v\nAll :%v\n\n", event, watcher.Add(f), watcher.WatchList())
        }
}
```

```
$ go build
$ ./foo &
[1] 516727
Starting wiht: [test]
$ echo a > test
WRITE         "test"
Add: <nil>
All :[test]

mv test{,bkp}
RENAME        "test"
Add: no such file or directory
All :[]

$ touch test
$ # nope, nothing. we are watching nothing. the CRL is lost now
```

---

To fix this, watch the CRL's parent directory, but only act upon the CRL file. Take this updated demo as an example.

```go
package main

import (
        "fmt"
        "path/filepath"

        "github.com/fsnotify/fsnotify"
)

func errCheck(err error) {
        if err != nil {
                panic(err)
        }
}

func main() {
        const f = "./test"

        watcher, err := fsnotify.NewWatcher()
        errCheck(err)
        defer func() { _ = watcher.Close() }()

        errCheck(watcher.Add(filepath.Dir(f)))
        fmt.Printf("Starting wiht: %v\n", watcher.WatchList())

        for event := range watcher.Events {
                if filepath.Base(event.Name) != filepath.Base(f) {
                        fmt.Printf("[SKIP] %+v\n", event)
                        continue
                }
                fmt.Printf("[ADD ] %+v\nAdd: %v\nAll :%v\n\n", event, watcher.Add(f), watcher.WatchList())
        }
}
```

```
$ go build
$ ./foo &
[2] 545169
Starting wiht: [.]

$ echo a > test
[ADD ] WRITE         "./test"
Add: <nil>
All :[. test]

$ mv test{,bkp}
[ADD ] RENAME        "./test"
Add: no such file or directory
All :[. test]

[SKIP] CREATE        "./testbkp" ← "./test"
[ADD ] RENAME        "test"
Add: no such file or directory
All :[.]

$ touch test
[ADD ] CREATE        "./test"
Add: <nil>
All :[. test]

[ADD ] CHMOD         "./test"
Add: <nil>
All :[. test]

```

config/tls.go
Jan-Schuppik: It's placed in TLSCommon rather than ServerTLS since CRL checking isn't inherently server-side, clients could use it as well. It hasn't been configured for client TLS yet, but keeping it common avoids duplicating the field/config plumbing if it's needed there.

config/tls.go
```diff
@@ -76,9 +91,95 @@ func (tc *TLSCommon) makeConfig() (*tls.Config, error) {
 		tlsConfig.Certificates = []tls.Certificate{crt}
 	}
 
+	if tc.CrlFile != "" {
+		info, err := os.Stat(tc.CrlFile)
+		if err != nil {
+			if os.IsNotExist(err) {
+				return nil, errors.New("CRL file doesn't exist")
+			}
+			return nil, errors.Wrap(err, "CRL file cannot be accessed")
```
oxzi: The error should be good enough by itself. There is no need to replace it with a custom one.


```suggestion
			return nil, errors.Wrap(err, "CRL file cannot be accessed")
```

config/tls.go
```diff
@@ -76,9 +91,95 @@ func (tc *TLSCommon) makeConfig() (*tls.Config, error) {
...
+		}
+		if info.IsDir() {
+			return nil, errors.New("CRL file is a directory")
+		}
```
oxzi: Why do you only check for directories? There are more options, such as sockets, links, and so on.


```suggestion
		if !info.Mode().IsRegular() {
			return nil, errors.New("CRL is not a regular file")
		}
```

config/crl.go
```diff
@@ -0,0 +1,160 @@
...
+
+	var signatureErr error
+	for _, issuer := range c.issuers {
+		if signatureErr = rl.CheckSignatureFrom(issuer); signatureErr == nil {
+			break
+		}
+	}
+	if signatureErr != nil {
+		return fmt.Errorf("CRL signature invalid: %w", signatureErr)
+	}
```
oxzi: For empty `c.issuers`, this would fail open. When reading this code, why do you even support multiple issuers and not one?

utils/crl.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+	}
+	if signatureErr != nil {
+		return fmt.Errorf("CRL signature invalid: %w", signatureErr)
+	}
+
+	revoked := make(map[string]struct{}, len(rl.RevokedCertificateEntries))
+	for _, entry := range rl.RevokedCertificateEntries {
+		revoked[entry.SerialNumber.Text(16)] = struct{}{}
+	}
```
oxzi: Of course, having a direct access cache might speed things up, but since you are allowing multiple issuers, you must distinguish those here. Otherwise, two certificates signed by different CAs with the identical serial number (yes, this happens, they are often just incrementing numbers) cannot be distinguished.

However, when creating `revoked`, you don't need most of `rl` in the `CrlChecker` struct anymore. Now you are storing duplicate data.

utils/crl.go
```diff
@@ -0,0 +1,160 @@
...
+	c.revoked = revoked
+	c.mu.Unlock()
+
+	return nil
+}
+
+// WatchAndReload blocks, watching the CRL file for changes and reloading it whenever a
+// Write, Create, or Rename event is received. It returns when ctx is canceled.
+// Reload failures are logged as warnings and do not stop the watcher. Callers typically
+// run this in a goroutine: go c.WatchAndReload(ctx, logger, nil).
```
oxzi: This "typically" example would ignore the error. This whole last sentence can be dropped, as it has no additional information.

utils/crl.go
```diff
@@ -0,0 +1,160 @@
...
+//
+// Atomic file replacements (e.g. mv tmp.crl ca.crl) are handled correctly across platforms:
+// on Linux/inotify the replacement fires Remove (IN_DELETE_SELF) on the old inode, while on
+// macOS/kqueue it fires Rename. The parent directory is watched so the watcher is not
+// confused by inode changes.
```
oxzi: This paragraph carries no information beyond internal reasoning, not relevant for a caller. Furthermore, it is outdated with the change from watching a single file to its directory.

utils/crl.go
```diff
@@ -0,0 +1,160 @@
...
+//
+// ready, if non-nil, is closed once the directory watch is installed and events will not be
+// missed. Callers that trigger CRL rotations programmatically (e.g. tests) should wait on
+// this channel before rotating the file.
+//
+// WatchAndReload returns an error only if the initial watcher setup fails; runtime errors are
```
oxzi: There are multiple returns in the for loop.

utils/crl.go
```diff
@@ -0,0 +1,160 @@
...
+// run this in a goroutine: go c.WatchAndReload(ctx, logger, nil).
+//
+// Atomic file replacements (e.g. mv tmp.crl ca.crl) are handled correctly across platforms:
+// on Linux/inotify the replacement fires Remove (IN_DELETE_SELF) on the old inode, while on
+// macOS/kqueue it fires Rename. The parent directory is watched so the watcher is not
+// confused by inode changes.
+//
+// ready, if non-nil, is closed once the directory watch is installed and events will not be
+// missed. Callers that trigger CRL rotations programmatically (e.g. tests) should wait on
+// this channel before rotating the file.
```
oxzi: Has this functionality any usage beyond testing? The code before closing the channel should be finished in less than a millisecond. I don't see any use for this.

config/crl.go
```diff
@@ -0,0 +1,160 @@
...
+				return fmt.Errorf("CRL watcher channel closed unexpectedly")
+			}
+
+			if filepath.Base(event.Name) != filepath.Base(c.path) {
+				continue
+			}
+
+			if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) || event.Has(fsnotify.Rename) {
+				if err := c.reload(); err != nil {
+					logger.Warnw("CRL reload failed", zap.Error(err))
```
oxzi, line 101: As you are only logging errors, a user will not be notified once the `CRLChecker` is healthy again. Consider adding a success message otherwise.

config/crl.go
```diff
@@ -0,0 +1,160 @@
...
+	if err := c.reload(); err != nil {
+		return nil, err
+	}
+	return c, nil
+}
+
+func (c *CrlChecker) reload() error {
+	data, err := os.ReadFile(c.path)
+	if err != nil {
+		if c.crl == nil {
```
oxzi: `c.crl` misses a read lock.

config/crl.go
```diff
@@ -0,0 +1,160 @@
...
+
+// IsRevoked reports whether serial appears in the CRL.
+// It returns an error if the CRL is outdated (NextUpdate in the past) or if the CRL cannot be accessed.
+func (c *CrlChecker) IsRevoked(serial *big.Int) (bool, error) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+
+	if c.crl == nil {
+		return false, fmt.Errorf("CRL is not loaded")
+	}
```
oxzi: How can this happen? In `NewCRLChecker`, `reload` is called before returning, populating the `crl` field. For later `reload` calls, the struct field is only populated in a success case, when `crl` is not `nil`.

utils/crl_test.go
```diff
@@ -0,0 +1,296 @@
...
+func atomicReplaceCRL(t *testing.T, path string, ca *x509.Certificate, caKey crypto.PrivateKey, nextUpdate time.Time, revokedSerials ...*big.Int) {
+	deleteCRLFile(path)
+	tmp, err := os.CreateTemp(filepath.Dir(path), "crl-new-*.pem")
+	require.NoError(t, err)
+	_, err = tmp.Write(buildCRL(t, ca, caKey, nextUpdate, revokedSerials...))
+	require.NoError(t, err)
+	require.NoError(t, tmp.Close())
+	require.NoError(t, os.Rename(tmp.Name(), path))
+
+	time.Sleep(200 * time.Microsecond)
```
oxzi: 200us is nowhere near enough as the reload runs asynchronous from this code. 200ms would be better.

```
$ go test -race -count=10 ./utils
--- FAIL: TestCRLChecker_WatchAndReload (2.01s)
    --- FAIL: TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement (0.00s)
        crl_test.go:226:
                Error Trace:    /home/eliza/git/github.com/Icinga/icinga-go-library/utils/crl_test.go:226
                Error:          Should be false
                Test:           TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement
--- FAIL: TestCRLChecker_WatchAndReload (2.01s)
    --- FAIL: TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement (0.00s)
        crl_test.go:226:
                Error Trace:    /home/eliza/git/github.com/Icinga/icinga-go-library/utils/crl_test.go:226
                Error:          Should be false
                Test:           TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement
--- FAIL: TestCRLChecker_WatchAndReload (1.91s)
    --- FAIL: TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement (0.00s)
        crl_test.go:226:
                Error Trace:    /home/eliza/git/github.com/Icinga/icinga-go-library/utils/crl_test.go:226
                Error:          Should be false
                Test:           TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement
--- FAIL: TestCRLChecker_WatchAndReload (2.01s)
    --- FAIL: TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement (0.00s)
        crl_test.go:226:
                Error Trace:    /home/eliza/git/github.com/Icinga/icinga-go-library/utils/crl_test.go:226
                Error:          Should be false
                Test:           TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement
--- FAIL: TestCRLChecker_WatchAndReload (2.01s)
    --- FAIL: TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement (0.00s)
        crl_test.go:226:
                Error Trace:    /home/eliza/git/github.com/Icinga/icinga-go-library/utils/crl_test.go:226
                Error:          Should be false
                Test:           TestCRLChecker_WatchAndReload/Reloads_on_atomic_file_replacement
FAIL
FAIL    github.com/icinga/icinga-go-library/utils       19.434s
FAIL
```

config/tls.go
```diff
@@ -76,9 +91,95 @@ func (tc *TLSCommon) makeConfig() (*tls.Config, error) {
...
+// rejects any client certificate whose serial number appears in the CRL. If a VerifyConnection
+// hook is already set on tlsConfig, the existing hook is called first and its error, if any,
+// takes precedence.
+//
+// The returned [utils.CrlChecker] owns the loaded CRL and must be kept alive for the duration
+// of the server's lifetime. Callers are responsible for calling [utils.CrlChecker.WatchAndReload]
+// to pick up CRL rotations at runtime.
+//
+// InitRevocationChecking is a no-op and returns (nil, nil) when tlsConfig is nil or when
+// either [TLSCommon.CrlFile] or [TLSCommon.Ca] is not configured.
```
oxzi: Why? Usually, when no error is returned, one can expect the real return argument to be non `nil`. This is only an invitation for panics.

config/tls.go
```diff
@@ -76,9 +91,95 @@ func (tc *TLSCommon) makeConfig() (*tls.Config, error) {
...
+func (tc *TLSCommon) InitRevocationChecking(tlsConfig *tls.Config) (*utils.CrlChecker, error) {
```
oxzi: I am repeating myself, but this whole logic makes no sense for `TLSCommon`, only for `ServerTLS`. Especially as `CrlFile` errors out in `MakeConfig`, but is only checked in `ServerTLS.Validate`. Please keep it clean and add move it to `ServerTLS`.

config/tls.go
```diff
@@ -35,6 +38,14 @@ type TLSCommon struct {
 	// or to verify client certificates in TLS server mode. Otherwise, the system's root CA pool is used.
 	// This option is ignored if Insecure is true in TLS client mode.
 	Ca string `yaml:"ca" env:"CA"`
+
+	// CrlFile is the path to the Certificate Revocation List (CRL) file.
+	//
+	// If specified, the CRL is used to check for revoked certificates in TLS.
+	// The CRL must be signed by the CA specified in the Ca option. If the CRL file is not found or cannot be loaded,
+	// an error is returned during TLS configuration.
+	// This option is ignored if TLS is not enabled or if the Ca option is not set.
```
oxzi: The `InitRevocationChecking` method should be mentioned, as only populating this field alone does not result in CRL checking.

utils/crl_test.go
```diff
@@ -0,0 +1,296 @@
...
+		template := &x509.RevocationList{
+			Number:     big.NewInt(1),
+			ThisUpdate: time.Now().Add(-time.Hour),
+			NextUpdate: future,
+		}
+		signer, ok := caKey.(crypto.Signer)
+		require.True(t, ok, "caKey must implement crypto.Signer")
+		der, err := x509.CreateRevocationList(rand.Reader, template, ca, signer)
+		require.NoError(t, err)
+		f, err := os.CreateTemp("", "crl-*.der")
```
oxzi: Above, you have used `t.TempDir()` instead of `""`. Please continue this pattern.

config/crl.go
oxzi, line 1: General question: Why did you place this in the `utils` package, left for helpers don't really fitting in anywhere else? I would consider to move this to `config`, next to the `tls.go` file.

config/tls.go
oxzi: > This option is ignored [...] if the Ca option is not set.

No, you check this in `hasCrlWithoutCa` and return an error.

config/tls.go
```diff
@@ -76,9 +91,95 @@ func (tc *TLSCommon) makeConfig() (*tls.Config, error) {
...
+		if err != nil {
+			if os.IsNotExist(err) {
+				return nil, errors.New("CRL file doesn't exist")
+			}
+			return nil, errors.Wrap(err, "CRL file cannot be accessed")
+		}
+		if info.IsDir() {
+			return nil, errors.New("CRL file is a directory")
+		}
+	}
```
oxzi: Honestly, this whole file verification is duplicate code you are also performing directly in the `CRLChecker` via `NewCRLChecker` (`reload`). Thus, every error raised here, would also be raised later in `InitRevocationChecking`. So consider just removing this, this is duplicate logic.

config/crl_test.go
```diff
@@ -0,0 +1,296 @@
...
+		IsCA:                  true,
+	}
+	der, err := x509.CreateCertificate(rand.Reader, template, template, key.Public(), key)
+	require.NoError(t, err)
+	cert, err := x509.ParseCertificate(der)
+	require.NoError(t, err)
+	return cert, key
+}
+
+func buildCRL(t *testing.T, ca *x509.Certificate, caKey crypto.PrivateKey, nextUpdate time.Time, revokedSerials ...*big.Int) []byte {
```
oxzi, line 46: `buildCRL` and `createCRLFile` are veeery similiar to `config.generateCRL`. When being in the same package - https://github.com/Icinga/icinga-go-library/pull/219/changes#r3585690885 -, they can be merged into one, reducing the test size.

config/tls.go
Jan-Schuppik: I overlooked IsRegular(), thanks!

config/crl.go
Jan-Schuppik: Oh i forgot the guard.
If I get it right, a CA file in PEM format can contain multiple certificates (a bundle), so InitRevocationChecking parses all of them and passes them along to find which CA actually signed the CRL.

utils/crl.go
Jan-Schuppik: Good point! I would say, I'll just drop the revoked cache, maintaining a separate lookup map feels like overkill for the performance gain.

utils/crl.go
Jan-Schuppik: I noticed the need during testing and thought it could be useful in general, but thinking about it more, every use case I could imagine feels fictional. So no real reason to keep it, I'll drop it.

config/crl.go
Jan-Schuppik: This can happen if the CrlChecker is created manually instead of with `NewCRLChecker`. Or should I not consider that case at all?

config/tls.go
Jan-Schuppik: All right, I have no objections

config/crl.go
Jan-Schuppik: I wasn't sure about the right package, so I started with utils, but config feels like the better fit.

config/crl.go
oxzi: In this case, all fields would be empty - including `path` - and no other method would work as well.

config/crl.go
```diff
@@ -0,0 +1,154 @@
...
+
+func (c *CrlChecker) reload() error {
+	data, err := os.ReadFile(c.path)
+	if err != nil {
+		c.mu.RLock()
+		if c.crl == nil {
+			// avoid nil pointer panic if CrlChecker is created without the factory method
+			return fmt.Errorf("cannot read initial CRL file %q: %w", c.path, err)
+		}
+		c.mu.RUnlock()
```
oxzi: Error case would return without releasing the lock.

config/crl.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}
+}
+
+// IsRevoked reports whether serial appears in the CRL.
+// It returns an error if the CRL is outdated (NextUpdate in the past) or if the CRL cannot be accessed.
+func (c *CrlChecker) IsRevoked(serial *big.Int) (bool, error) {
+	if c == nil {
+		return false, fmt.Errorf("crl is required")
+	}
```
oxzi: Can be removed. There is no valid path for `c` to be `nil`. If this has happened, something else went wrong horribly before.

config/crl.go
oxzi: My prior comment still stands: There is no valid path how this can happen.

config/crl_test.go
```diff
@@ -0,0 +1,295 @@
...
+	require.NoError(t, err)
+	require.NoError(t, tmp.Close())
+	require.NoError(t, os.Rename(tmp.Name(), path))
+
+	time.Sleep(200 * time.Millisecond)
+}
+
+func deleteCRLFile(path string) {
+	_ = os.Remove(path)
+}
```
oxzi: This helper function adds no surplus value. Just call `os.Remove` directly.

config/crl.go
```diff
@@ -0,0 +1,154 @@
...
+	path    string
+	issuers []*x509.Certificate
+	mu      sync.RWMutex
+	crl     *x509.RevocationList
+}
+
+// NewCRLChecker creates a [CrlChecker] by loading the CRL at path and verifying its signature
+// against issuer. The file may be PEM- or DER-encoded. An error is returned if the file cannot
+// be read, is not a valid CRL, or its signature does not match issuer.
+func NewCRLChecker(path string, issuers ...*x509.Certificate) (*CrlChecker, error) {
```
oxzi: You are still allowing multiple issuers, but doing a comparison based on the serial number.

https://github.com/Icinga/icinga-go-library/blob/9c7f640cf598527c7ce39815648c47aea7ff8bdd/config/tls.go#L327

This is the same issue I [mentioned last time](https://github.com/Icinga/icinga-go-library/pull/219#discussion_r3585395015), just from another angle.

---

So, to keep it simple, I would argue to only allow one single CA. If someone configures a CA bundle with either less or more than one entry, this is an error. For this setup, this is totally fine, especially when documented with an additional sentence in the docs.

config/crl.go
oxzi: Would be addressed with the suggestion in <https://github.com/Icinga/icinga-go-library/pull/219/changes#r3613899290>.

config/crl.go
```diff
@@ -0,0 +1,154 @@
...
+		c.mu.RLock()
+		if c.crl == nil {
+			// avoid nil pointer panic if CrlChecker is created without the factory method
+			return fmt.Errorf("cannot read initial CRL file %q: %w", c.path, err)
+		}
+		c.mu.RUnlock()
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("cannot read CRL file %q: %w", c.path, err)
+		}
+		return nil
```
oxzi: Why ignore `fs.ErrNotExist`? Couldn't this be simplified to:


```suggestion
		return fmt.Errorf("cannot read CRL file %q: %w", c.path, err)
```

---

Btw, `os.IsNotExist`'s API doc states that this function should not be used for new code, but `errors.Is(err, fs.ErrNotExist)` instead.

config/tls.go
```diff
@@ -243,3 +263,76 @@ func (st *ServerTLS) MakeConfig() (*tls.Config, error) {
...
+// rejects any client certificate whose serial number appears in the CRL. If a VerifyConnection
+// hook is already set on tlsConfig, the existing hook is called first and its error, if any,
+// takes precedence.
+//
+// The returned [utils.CrlChecker] owns the loaded CRL and must be kept alive for the duration
+// of the server's lifetime. Callers are responsible for calling [utils.CrlChecker.WatchAndReload]
+// to pick up CRL rotations at runtime.
+//
+// InitRevocationChecking is a no-op and returns a no-op-crlChecker when tlsConfig is nil or when [TLSCommon.CrlFile]
+// is empty. It returns an error if the CA certificate or CRL cannot be loaded or verified.
```
oxzi: This is outdated, right?

config/tls.go
```diff
@@ -243,3 +263,76 @@ func (st *ServerTLS) MakeConfig() (*tls.Config, error) {
...
+//
+// It loads the CA certificate from [TLSCommon.Ca] and the CRL from [TLSCommon.CrlFile], verifies
+// the CRL's signature against the CA, and installs a [tls.Config.VerifyConnection] hook that
+// rejects any client certificate whose serial number appears in the CRL. If a VerifyConnection
+// hook is already set on tlsConfig, the existing hook is called first and its error, if any,
+// takes precedence.
+//
+// The returned [utils.CrlChecker] owns the loaded CRL and must be kept alive for the duration
+// of the server's lifetime. Callers are responsible for calling [utils.CrlChecker.WatchAndReload]
+// to pick up CRL rotations at runtime.
```
oxzi, line 279: The links are outdated. There is no `utils` package anymore.

config/tls.go
```diff
@@ -217,6 +234,9 @@ func (st *ServerTLS) Validate() error {
 // If TLS is not enabled (st.Enable is false), it returns nil without an error.
 func (st *ServerTLS) MakeConfig() (*tls.Config, error) {
 	tlsConfig, err := st.makeConfig()
+	if st.CrlFile != "" && st.Ca == "" {
+		return nil, errors.New("CRL file given, but CA certificate missing")
+	}
```
oxzi, line 240: This should either be moved above `st.makeConfig` as it now sits between the method call and the `err` checking.

oxzi (commented):
Please rebase the PR branch on top of the `main` branch. While there are no merge conflicts, the branches are slowly diverging. Furthermore, please squash the commits into one.

config/crl.go
oxzi: Pardon, but this is still there.

config/crl.go
```diff
@@ -0,0 +1,152 @@
...
+	rl, err := x509.ParseRevocationList(data)
+	if err != nil {
+		return fmt.Errorf("cannot parse CRL: %w", err)
+	}
+
+	if len(c.issuers) == 0 {
+		return fmt.Errorf("no issuer certificates provided to verify CRL signature")
+	} else if len(c.issuers) > 1 {
+		return fmt.Errorf("multiple issuer certificates provided; expected exactly one")
+	}
```
oxzi: How can this happen? You are only setting `c.issuers` via `NewCRLChecker`, where an identical check already exists.

Comment would be void with patch from https://github.com/Icinga/icinga-go-library/pull/219/changes#r3620350869

config/crl.go
```diff
@@ -0,0 +1,152 @@
...
+	"go.uber.org/zap"
+)
+
+// CrlChecker loads a Certificate Revocation List (CRL) from a file, verifies its signature
+// against a trusted CA certificate, and exposes [CrlChecker.IsRevoked] for serial number
+// lookups. The in-memory CRL is protected by a read/write mutex so that background reloads
+// via [CrlChecker.WatchAndReload] are safe to run concurrently with ongoing checks.
+type CrlChecker struct {
+	path    string
+	issuers []*x509.Certificate
```
oxzi: Since we have agreed on one single CA certificate, there is no need to store them as an array. Reflecting this in the code also removes lines and makes it easier to read.

```diff
diff --git a/config/crl.go b/config/crl.go
index b535c13..718a31f 100644
--- a/config/crl.go
+++ b/config/crl.go
@@ -22,24 +22,17 @@ import (
 // lookups. The in-memory CRL is protected by a read/write mutex so that background reloads
 // via [CrlChecker.WatchAndReload] are safe to run concurrently with ongoing checks.
 type CrlChecker struct {
-       path    string
-       issuers []*x509.Certificate
-       mu      sync.RWMutex
-       crl     *x509.RevocationList
+       path   string
+       issuer *x509.Certificate
+       mu     sync.RWMutex
+       crl    *x509.RevocationList
 }

 // NewCRLChecker creates a [CrlChecker] by loading the CRL at path and verifying its signature
 // against issuer. The file may be PEM- or DER-encoded. An error is returned if the file cannot
 // be read, is not a valid CRL, or its signature does not match issuer.
-//
-// If a CA bundle is used to sign the CRL, it must contain exactly one certificate, the issuer of the CRL.
-// Multiple certificates in the bundle will result in an error.
-func NewCRLChecker(path string, issuers ...*x509.Certificate) (*CrlChecker, error) {
-       if len(issuers) != 1 {
-               return nil, fmt.Errorf("expected exactly one issuer certificate, got %d", len(issuers))
-       }
-
-       c := &CrlChecker{path: path, issuers: issuers}
+func NewCRLChecker(path string, issuer *x509.Certificate) (*CrlChecker, error) {
+       c := &CrlChecker{path: path, issuer: issuer}
        if err := c.reload(); err != nil {
                return nil, err
        }
@@ -69,13 +62,7 @@ func (c *CrlChecker) reload() error {
                return fmt.Errorf("cannot parse CRL: %w", err)
        }

-       if len(c.issuers) == 0 {
-               return fmt.Errorf("no issuer certificates provided to verify CRL signature")
-       } else if len(c.issuers) > 1 {
-               return fmt.Errorf("multiple issuer certificates provided; expected exactly one")
-       }
-
-       if signatureErr := rl.CheckSignatureFrom(c.issuers[0]); signatureErr != nil {
+       if signatureErr := rl.CheckSignatureFrom(c.issuer); signatureErr != nil {
                return fmt.Errorf("CRL signature invalid: %w", signatureErr)
        }

diff --git a/config/crl_test.go b/config/crl_test.go
index 741c736..bca63ff 100644
--- a/config/crl_test.go
+++ b/config/crl_test.go
@@ -147,11 +147,6 @@ func TestNewCRLChecker(t *testing.T) {
                _, err := NewCRLChecker(path, ca)
                require.ErrorContains(t, err, "CRL signature invalid")
        })
-
-       t.Run("Multiple CA certificates", func(t *testing.T) {
-               _, err := NewCRLChecker("/nonexistent/crl.pem", ca, ca)
-               require.ErrorContains(t, err, "expected exactly one issuer certificate")
-       })
 }

 func TestCRLChecker_IsRevoked(t *testing.T) {
diff --git a/config/tls.go b/config/tls.go
index 7d2fc2b..264778e 100644
--- a/config/tls.go
+++ b/config/tls.go
@@ -301,11 +301,11 @@ func (st *ServerTLS) InitRevocationChecking(tlsConfig *tls.Config) (*CrlChecker,
                }
                caCerts = append(caCerts, cert)
        }
-       if len(caCerts) == 0 {
-               return nil, errors.New("failed to decode X.509 certificate")
+       if len(caCerts) != 1 {
+               return nil, fmt.Errorf("expected one CA certificate in the CA bundle, got %d", len(caCerts))
        }

-       checker, err := NewCRLChecker(st.CrlFile, caCerts...)
+       checker, err := NewCRLChecker(st.CrlFile, caCerts[0])
        if err != nil {
                return nil, fmt.Errorf("cannot load CRL: %w", err)
        }
```

config/tls.go
```diff
@@ -188,6 +190,18 @@ type ServerTLS struct {
...
+
+	// CrlFile is the path to the Certificate Revocation List (CRL) file.
+	//
+	// If specified, the CRL is used to check for revoked certificates in TLS.
+	// The CRL must be signed by the CA specified in the Ca option. If the CRL file is not found or cannot be loaded,
+	// an error is returned during TLS configuration.
+	// This option is ignored if TLS is not enabled. If the Ca option is not set it returns an error.
+	//
+	// Note that populating this field alone does not enable CRL checking.
+	// [TLSCommon.InitRevocationChecking] must be called after assembling the [tls.Config]
```
oxzi:
```suggestion
	// [ServerTLS.InitRevocationChecking] must be called after assembling the [tls.Config]
```

Moved down.

config/crl.go
```diff
@@ -0,0 +1,152 @@
...
+	mu      sync.RWMutex
+	crl     *x509.RevocationList
```
oxzi: With switching to a single CA architecture, we can easily use a map here, as you had in a prior iteration.

oxzi (commented):
After these changes, all should be fine.

Btw, I have pushed a changed version addressing (most of) my review to the [`add-client-cert-revoke-check-review-apenning` branch](https://github.com/Icinga/icinga-go-library/tree/add-client-cert-revoke-check-review-apenning). Feel free to absorb it.

oxzi (approved):

PR #222, opened 2026-07-15
notifications: introduce `YieldIncidents` function

yhabteab:
As an alternative to the existing `GetIncidents` function which reads and decodes the entire response body before returning the huge slice of incidents. This new function, on the other hand, reads, decodes and yields a single `source.Incident` object at a time until the EOF.

### DIFF
--- a/notifications/source/client.go
+++ b/notifications/source/client.go
@@ -18,6 +18,17 @@ import (
 	"github.com/pkg/errors"
 )
 
+var (
+	// ErrAttrsNegotiation implies missing attributes.
+	ErrAttrsNegotiation = stderrors.New("attribute negotiation required")
+
+	// ErrUnauthorizedRequest indicates that the request was unauthorized, typically due to invalid credentials.
+	ErrUnauthorizedRequest = stderrors.New("unauthorized request")
+
+	// ErrReadPartialResp indicates that a partial response was received from the API, which may require special handling.
+	ErrReadPartialResp = stderrors.New("read partial response")
+)
+
 // clientTransport is a http.RoundTripper to be used in NewClient.
 type clientTransport struct {
 	http.RoundTripper
@@ -111,9 +122,6 @@ func NewClient(cfg Config, clientName string) (*Client, error) {
 	}, nil
 }
 
-// ErrAttrsNegotiation implies missing attributes.
-var ErrAttrsNegotiation = stderrors.New("attribute negotiation required")
-
 // ProcessEvent submits an event.Event to the Icinga Notifications /process-event API endpoint.
 //
 // When rejectIncomplete is set and the returned error is ErrAttrsNegotiation,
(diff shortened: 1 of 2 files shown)


notifications/source/client.go
```diff
@@ -143,34 +153,101 @@ func (client *Client) ProcessEvent(ctx context.Context, ev *event.Event, rejectI
...
-	return incidents, nil
+		//nolint:bodyclose // False positive, drainBody is called in the defer statement below.
+		resp, err := client.doRequest(ctx, http.MethodGet, client.endpoints.Incidents, filter, nil, nil)
+		if err != nil {
+			errCh <- errors.Wrap(err, "cannot GET incidents from API")
+			return
+		}
+		defer drainBody(resp.Body)
+
+		if resp.StatusCode != http.StatusAccepted {
```
oxzi, line 229: At the moment, this should return 200, as the code did before this diff.

https://github.com/Icinga/icinga-notifications/blob/6158b8a73a6cbddb82fbea4443af25fc3fedc7e7/internal/listener/listener.go#L600-L607

notifications/source/client.go
```diff
@@ -201,16 +278,27 @@ func (client *Client) ModifyIncidents(ctx context.Context, attrs ModifiableIncid
 	}
 	defer drainBody(resp.Body)
 
-	if resp.StatusCode != http.StatusOK {
+	if resp.StatusCode != http.StatusAccepted {
```
oxzi, line 298: Same as https://github.com/Icinga/icinga-go-library/pull/222/changes#r3602836938 - are you planning to change this in another PR or did I miss something?

notifications/source/client.go
```diff
@@ -259,11 +347,37 @@ func (client *Client) doRequest(
...
+}
+
+// String returns a string representation of the IncidentStatus, including the code and status message.
+func (s IncidentStatus) String() string {
+	if s.Code < 100 || s.Code > 599 || s.Status == "" {
+		return ""
+	}
+	if s.Status == "" {
+		return fmt.Sprintf("code %d", s.Code)
+	}
```
oxzi: The second `if` can never be reached, as you are already checking `s.Status == ""` in the first. The `return ""` brings the loop in `YieldIncidents` to abort.

notifications/source/client.go
```diff
@@ -259,11 +347,37 @@ func (client *Client) doRequest(
...
+//
+// All other fields are set according to the incident's current state at the time of the request.
 type Incident struct {
 	IsMuted    bool              `json:"is_muted"`
-	ObjectTags map[string]string `json:"object_tags"`
-	Severity   event.Severity    `json:"severity"`
+	ObjectTags map[string]string `json:"object_tags,omitempty"`
+	Severity   event.Severity    `json:"severity,omitempty"`
+
+	IncidentStatus `json:",inline"`
```
oxzi: There is no `,inline` for `json`, this is `yaml`. Same below in `ModifiedIncidentResp`.

oxzi (commented):
Here is a first review of this PR. I know, you haven't requested one, but the PR was marked as ready and I had a bit of spare time. I guess, it still depends on future changes, right?

notifications/source/client.go
yhabteab: Icinga DB uses that everywhere but yeah, the JSON encoder/decoder automatically inline embedded structs.

notifications/source/client.go
yhabteab: That will change with an upcoming PR in Icinga Notifications.

yhabteab:
> I guess, it still depends on future changes, right?

Yep. It depends on this one:
- https://github.com/Icinga/icinga-notifications/pull/471

notifications/source/client.go
```diff
@@ -286,11 +374,37 @@ func (client *Client) doRequest(
...
+// In that case, such an object will be the last object in the response stream and the client should handle it
+// accordingly. In all other cases, the Status field will be zero and can be ignored.
+//
+// All other fields are set according to the incident's current state at the time of the request.
 type Incident struct {
 	IsMuted    bool              `json:"is_muted"`
-	ObjectTags map[string]string `json:"object_tags"`
-	Severity   event.Severity    `json:"severity"`
+	ObjectTags map[string]string `json:"object_tags,omitempty"`
+	Severity   event.Severity    `json:"severity,omitempty"`
```
oxzi, line 383: Out of curiosity: Why did you add the `omitempty` struct tag here? Is it just for completeness with the other structs? Unless I missed something, all `IncidentStatus`, `Incident` and `ModifiedIncidentResp` are only being used for unmarshalling, making the `omitempty` a noop.

notifications/source/client.go
```diff
@@ -170,34 +180,101 @@ func (client *Client) ProcessEvent(ctx context.Context, ev *event.Event, rejectI
...
+				return ctx.Err()
+			}
+		}
+		return nil
+	})
+	if err := g.Wait(); err != nil {
+		return nil, err
 	}
-	defer drainBody(resp.Body)
+	return incidents, nil
```
oxzi: Taking a look at `YieldIncidents`, I don't think you need the `errgroup` dance here. Since `errCh` is being buffered and `YieldIncidents` honors the same `ctx` for each incident in its loop, I think all this can be reduced to something like:


```suggestion
	for incident := range incidentsCh {
		incidents = append(incidents, incident)
	}
	if err := <- errCh; err != nil {
		return nil, err
	}
	return incidents, nil
```

oxzi (commented):
In general, looks good. Thanks.

However, I would wait with this PR for Icinga/icinga-notifications#471 as I suggested some JSON field changes over there.

In case you are overly motivated, you could consider extending the `client_test.go` I recently added and return some static NDJSON there and call `GetIncidents` instead of `ProcessEvent`.

notifications/source/client.go
yhabteab: Icinga Notifications uses that very same struct to encode its responses as well.

notifications/source/client.go
yhabteab: Yeah, looks better that way. Ty.

yhabteab:
> In case you are overly motivated, you could consider extending the `client_test.go` I recently added and return some static NDJSON there and call `GetIncidents` instead of `ProcessEvent`.

Overly motivated?? I've added some basic test cases for the two endpoints.

notifications/source/client.go
oxzi: Oh, yeah. Seems like I had 🍅 on my 👀 .

notifications/source/client_test.go
```diff
@@ -107,6 +126,90 @@ func TestClient(t *testing.T) {
...
+			handler: func(t *testing.T, rw http.ResponseWriter, r *http.Request) bool {
+				incidents := []any{
+					Incident{IsMuted: false, ObjectTags: map[string]string{"icinga": "test"}, Severity: event.SeverityCrit},
+					Incident{IsMuted: true, ObjectTags: map[string]string{"icinga": "test2"}, Severity: event.SeverityWarning},
+					Incident{IsMuted: false, ObjectTags: map[string]string{"icinga": "test3"}, Severity: event.SeverityInfo},
+				}
+				writeResp(t, rw, incidents)
+				return true
+			},
+			conf: func(srv *httptest.Server) Config { return Config{Url: "http://" + srv.Listener.Addr().String()} },
```
oxzi:
```suggestion
			conf: func(srv *httptest.Server) Config { return Config{Url: srv.URL} },
```

Same for the following tests.

notifications/source/client.go
```diff
@@ -286,11 +364,25 @@ func (client *Client) doRequest(
...
 }
 
+// ErrorState represents the error state of an incident as returned by the Icinga Notifications /incidents API endpoint.
+type ErrorState struct {
+	Error string `json:"error,omitempty"`
+}
+
 // Incident represents a single incident object as returned by the Icinga Notifications /incidents API endpoint.
+//
+// Since Icinga Notifications streams the response in a NDJSON format, the Status field will be set to a non-zero
```
oxzi: There is no status field anymore.

notifications/source/client.go
```diff
@@ -334,3 +437,16 @@ func drainBody(body io.ReadCloser) {
 	_, _ = io.Copy(io.Discard, body)
 	_ = body.Close()
 }
+
+// ModifyError represents an error that occurred while modifying incidents via the /incident endpoint.
+type ModifyError struct {
+	results []ModifiedIncidentResp
+}
+
+// Error implements the error interface for ModifyIncidentError.
```
oxzi:
```suggestion
// Error implements the error interface for ModifyError.
```

Right?

yhabteab:
Addressed the requested changes and also enabled parallel testing for the client test cases as well.

yhabteab:
> also enabled parallel testing for the client test cases as well.

Now, it's fully enabled.

oxzi (approved):
LGTM, thanks!

PR #224, opened 2026-07-16
notifications.source: Unix Domain Socket and HTTPS

oxzi:
Support both Unix Domain Sockets and HTTPS as alternative transports for the notifications source client. Based on the URL's scheme, a transport is chosen and different constraints are applied for the configuration.

---

So far, I have tested this change with Icinga DB - where I bumped the IGL to this branch - and Icinga Notifications either at the `main` branch for the Unix Domain Socket and HTTPS w/ Basic Auth or at Icinga/icinga-notifications#454 for TLS Client Certificates.

There, I realized that Go's HTTP module is more picky than `curl(1)` regarding server certificates, expecting a SAN. Thus, for creating certs adjust my [prior comment](https://github.com/Icinga/icinga-notifications/pull/454#pullrequestreview-4633722749) with this modification for the server certificate.

```
$ cat server_cert.ext
subjectKeyIdentifier   = hash
authorityKeyIdentifier = keyid:always,issuer:always
basicConstraints       = CA:TRUE
keyUsage               = digitalSignature, nonRepudiation, keyEncipherment, dataEncipherment, keyAgreement, keyCertSign
subjectAltName         = DNS:example.com, DNS:*.example.com
issuerAltName          = issuer:copy

$ openssl x509 -req -days 1 -in server.csr -CA ca.crt -CAkey ca.key -CAcreateserial -extfile server_cert.ext -out server.crt
```

Refs Icinga/icingadb#1113.

### DIFF
--- a/notifications/source/config.go
+++ b/notifications/source/config.go
@@ -1,20 +1,43 @@
 package source
 
 import (
+	"net/url"
+
 	"github.com/icinga/icinga-go-library/config"
+	"github.com/pkg/errors"
+)
+
+const (
+	// Supported URI schemes in Config.Url.
+	schemeHttp  = "http"
+	schemeHttps = "https"
+	schemeUnix  = "unix"
 )
 
 // Config defines all configuration for the Icinga Notifications API Client.
 type Config struct {
-	// Url points to the Icinga Notifications API, e.g., http://localhost:5680
+	// Url of the Icinga Notifications API.
+	//
+	// A transport is chosen based on the URI scheme:
+	//   - http: Unencrypted HTTP connection. Requires Username and Password/PasswordFile to be set.
+	//     For example: http://example.com:5680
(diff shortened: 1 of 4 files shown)


notifications/source/config.go
```diff
@@ -1,29 +1,95 @@
...
+			return err
+		}
+		if c.Username == "" || c.Password == "" {
+			return errors.New("http notifications source requires a username and password")
+		}
+
+	case schemeHttps:
+		c.TlsOptions.Enable = true
+		if c.TlsOptions.Cert == "" && c.Username == "" {
+			return errors.New("https notifications source either requires certificates or username and password")
```
yhabteab:
```suggestion
			return errors.New("https notifications source requires either certificates or username and password")
```

Don't forget the test case as well.

yhabteab (commented):
Wow, you've even added HTTP cleint test cases :). Apart from the below small change, looks fine. I didn't test it though but reading the PR description, you seem to aleady have done that, so all good.

yhabteab (approved):

PR #226, opened 2026-07-27
notifications: replace `rpc` with real `jsonrpc` package

yhabteab:
This PR replaces the Icinga Notifcations `rpc` package with a real `jsonrpc` package. The `jsonrpc` package provides a more robust and standardized implementation of the JSON-RPC protocol, which allows for better interoperability and easier maintenance of the codebase. The `rpc` package was a very basic implementation of the JSON-RPC protocol without fully supporting all the features and specifications of the protocol. As a result, each plugin developer using a language other than Go was doomed to implement their own JSON-RPC server that is compatible with the Icinga Notifications `rpc` package, instead of just using a standard JSON-RPC implementation of their choice. This PR aims to address this issue by using a well-established and fairly simple [`jsonrpc2`](https://github.com/sourcegraph/jsonrpc2) package that adheres to the [JSON-RPC 2 specifications](https://www.jsonrpc.org/specification), making it easier for developers to create compatible plugins without having to worry about the underlying implementation details of the `rpc` package.

Also, the `jsonrpc` package allows plugins to perform RPC calls back to the Icinga Notifications beyond the simple request-response model of the `rpc` package. This allows plugins, for instance, to call the `Log` method of the Icinga Notifications (that's going to be implemented in a future PR) to log messages with the desired log level.


notifications/jsonrpc/debug_off.go
oxzi, line 1: Please consider renaming the `debug.go` and `nodebug.go` files to something which appears sequentially in the file listing. Otherwise, when opening this package for the first time (or after a time, again), this is a bit confusing. Dunno, `jsonrpc_debug_on.go` and `jsonrpc_debug_off.go` or something.

notifications/jsonrpc/jsonrpc.go
```diff
@@ -0,0 +1,159 @@
...
+	w io.WriteCloser
+}
+
+func (rwc *readWriteCloser) Read(p []byte) (n int, err error)  { return rwc.r.Read(p) }
+func (rwc *readWriteCloser) Write(p []byte) (n int, err error) { return rwc.w.Write(p) }
+func (rwc *readWriteCloser) Close() error {
+	if err := rwc.r.Close(); err != nil {
+		return err
+	}
+	return rwc.w.Close()
```
oxzi: This would skip closing the writer. How about:

```suggestion
	return errors.Join(wc.r.Close(), rwc.w.Close())
```

notifications/jsonrpc/jsonrpc.go
```diff
@@ -0,0 +1,159 @@
...
+		jsonrpc2.OnSend(onJsonRpcSend),
+		jsonrpc2.OnRecv(onJsonRpcRecv),
+	)
+	return ep
+}
+
+// Conn returns the underlying JSON-RPC connection.
+func (e *Endpoint) Conn() *Conn { return e.conn }
+
+// Done returns a channel that is closed when the underlying connection is closed or the context is canceled.
```
oxzi, line 128: > or the context is canceled.

The context part does not check out. The API doc comment of `Conn.DisconnectNotify` only mentions the underlying connection.

Furthermore, I have just traced the context given to `jsonrpc2.NewConn`. From there, it goes to `go c.readMessage(ctx)`, and is only passed further to the handler. The `Conn.readMessage` method does not honor the context. In your tests, this works as you are manually closing the internal `jsonrpc2.Conn`. Fortunately, the same applies to the `Run` function.

https://github.com/Icinga/icinga-go-library/blob/ab6a2b6fa468d0f13598e6d9568067bc2d264614/notifications/jsonrpc/jsonrpc_test.go#L105-L108

notifications/jsonrpc/jsonrpc.go
```diff
@@ -0,0 +1,159 @@
...
+// However, the plugins are expected to pass in their respective stdin and stdout streams for communication with
+// Icinga Notifications. When done, the caller is responsible for closing the RPC connection by calling the
+// [Conn.Close] method on the returned endpoint.
+func New(ctx context.Context, r io.ReadCloser, w io.WriteCloser, h Handler) *Endpoint {
+	ep := new(Endpoint)
+	ep.conn = jsonrpc2.NewConn(
+		ctx,
+		jsonrpc2.NewPlainObjectStream(&readWriteCloser{r, w}),
+		h,
+		jsonrpc2.SetLogger(loggerFn(ep.NotifyLog)),
```
oxzi: The API docs for `SetLogger` states that it "sets the logger for the connection". But now we are using this very underlying connection for the logger itself. I traced this a bit and it seems that this can result in a deadlock.

When the underlying `Conn` is closed with a known cause, the logger is called while the internal mutex is locked.

https://github.com/sourcegraph/jsonrpc2/blob/v0.2.1/conn.go#L169-L182

Now, `c.logger.Printf` results in `e.conn.Notify` being called, which calls `c.send`.

https://github.com/sourcegraph/jsonrpc2/blob/v0.2.1/conn.go#L127-L142

And there, another mutex lock is being set, while the prior lock is still active.

https://github.com/sourcegraph/jsonrpc2/blob/v0.2.1/conn.go#L241-L260

Btw, `Conn.close` is only called with a cause in `c.readMessage` if the next JSON RPC message cannot be read. Thus, going back to Icinga Notifications, an incorrect (or malicious) channel plugin might cause a deadlock in Icinga Notifications.

As this sounds like an API footgun, I have reported it upstream: https://github.com/sourcegraph/jsonrpc2/issues/98

notifications/plugin/plugin.go
```diff
@@ -215,69 +228,90 @@ func PopulateDefaults(typePtr Plugin) error {
...
+
+	case MethodSetConfig:
+		if req.Params == nil {
+			if err := jsonrpc.ReplyMissingParams(ctx, conn, req.ID); err != nil {
+				log.Fatalf("failed to send SetConfig response: %v", err)
 			}
+		}
 
-			log.Fatal("failed to read request:", err)
+		if err := rh.p.SetConfig(*req.Params); err != nil {
```
oxzi, line 254: The `req.Params` nil check above this if does only abort if `jsonrpc.ReplyMissingParams` fails, but would continue otherwise, resulting in a nil pointer dereferencing here.

Second thoughts: Every error path in `MethodSetConfig` and `MethodSendNotification` misses a return. In the end, either multiple `ReplyMissing`/`ReplyError` calls and a final `Reply` calls are made.

notifications/plugin/plugin.go
```diff
@@ -215,69 +228,90 @@ func PopulateDefaults(typePtr Plugin) error {
...
+// of the Plugin interface (see [Run]).
+type rpcHandler struct {
+	p Plugin
+}
+
+func (rh *rpcHandler) Handle(ctx context.Context, conn *jsonrpc.Conn, req *jsonrpc.Request) {
+	switch req.Method {
+	case MethodGetInfo:
+		if err := conn.Reply(ctx, req.ID, rh.p.GetInfo()); err != nil {
+			log.Fatalf("failed to send GetInfo response: %v", err)
```
oxzi, line 243: While the old code already `panic`ed and a fatal log does not change this behavior, this is not documented. Thus, either document that a `Run` call might result in process termination or consider changing this to an returned error. (I am totally fine with `log.Fatal` as long as it is documented.)

notifications/jsonrpc/jsonrpc.go
oxzi: I have only skimmed the daemon code in the Icinga Notifications PR so far, but wouldn't this logger result in also sending log messages from the Icinga Notifications daemon to the channel plugin? Do we want this?

notifications/plugin/plugin.go
yhabteab: 🐘

notifications/jsonrpc/jsonrpc.go
yhabteab: > Do we want this?

Nope, just didn't think of that :(.

yhabteab: Oh gosh, thank you for digging this out before landing in production. Obviously, I didn't trace it that deep to spot such a severe _bug (or I just used the library horribly wrong)_. I've now changed the implementation to send the notifications from within a different goroutine.

notifications/jsonrpc/jsonrpc.go
```diff
@@ -0,0 +1,192 @@
...
+func New(ctx context.Context, r io.ReadCloser, w io.WriteCloser, h Handler, reportRPCErrorsToPeer bool) *Endpoint {
+	opts := []jsonrpc2.ConnOpt{
+		jsonrpc2.OnSend(onJsonRpcSend),
+		jsonrpc2.OnRecv(onJsonRpcRecv),
+	}
+	ep := new(Endpoint)
+	if reportRPCErrorsToPeer {
+		opts = append(opts, jsonrpc2.SetLogger(loggerFn(ep.NotifyLog)))
+		ep.logsCh = make(chan *LogParams)
+	}
```
oxzi: If `!reportRPCErrorsToPeer`, the generic `log` is used, creating log messages not matching our zap loggers. How about passing an zap logger otherwise?

https://github.com/sourcegraph/jsonrpc2/blob/v0.2.1/conn.go#L54

notifications/jsonrpc/jsonrpc.go
```diff
@@ -0,0 +1,192 @@
...
+	if e.logsCh == nil {
+		return fmt.Errorf("NotifyLog called on an endpoint that doesn't support logging")
+	}
+
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case e.logsCh <- &LogParams{Level: lvl, Message: msg, Fields: fields}:
+		return nil
+	}
```
oxzi, line 154: It seems there is another aspect of https://github.com/Icinga/icinga-go-library/pull/226#discussion_r3673116372. Starting with the (non-) obvious: The `Conn.logger` must not block, as otherwise `Conn.close` will be blocked as well.

https://github.com/sourcegraph/jsonrpc2/blob/v0.2.1/conn.go#L180-L186

---

1. The `ctx.Done()` exit case does not work as expected, as its context is `context.Background()`; not the same context passed to `Endpoint.serveLogNotifications` via `New`. And its `Done()` implementation returns a `nil` channel.

   - https://github.com/Icinga/icinga-go-library/blob/07122f2f2fef0ab442f2d055dc0545c32fdc919d/notifications/jsonrpc/jsonrpc.go#L191
   - https://cs.opensource.google/go/go/+/refs/tags/go1.26.5:src/context/context.go;l=187-199
  
   So, at least the API doc regarding the context cancellation is wrong.
   https://github.com/Icinga/icinga-go-library/blob/07122f2f2fef0ab442f2d055dc0545c32fdc919d/notifications/jsonrpc/jsonrpc.go#L127-L128

2. The `logsCh` consumer has another exit case missing here: `e.Done`, wrapping `e.conn.DisconnectNotify` and returning the `Conn.disconnect` channel closed in `Conn.close`.
   https://github.com/sourcegraph/jsonrpc2/blob/v0.2.1/conn.go#L180-L184

3. While `Endpoint.serveLogNotifications` drains `e.logsCh` in another goroutine, it uses the same internal `Conn.mu` mutex. Honestly, I am not quite sure at the moment, but I would guess that it can happen that internal `Conn.send` calls via `Endpoint.NotifyLog` and `Conn.close` calls might race the mutex and block, due to the unbuffered `logsCh`.

---

To not go too far, let's consider the scope of this potential error: This should only happen within channel plugins, not within Icinga Notifications itself. And it might happen, that under certain circumstances, the `jsonrpc2.Conn` might block or even deadlock. However, this should only happen in a fatal error case, where the channel plugin should die on anyway.

So, how about storing the context passed to `New` in the `loggerFn` type as well - either by making it a struct or through a closure -, and make `logsCh` buffered - dunno, 64 entries - and add a `default` case to bail out even if the buffered channel is full.

notifications/plugin/plugin.go
```diff
@@ -196,6 +201,14 @@ type Plugin interface {
 	SendNotification(req *NotificationRequest) error
 }
 
+// RPCEndpointReceiver is an interface that can be implemented by a plugin to receive the RPC endpoint used internally.
+//
+// If a channel plugin implements this interface, the [Run] function will inject the RPC endpoint into the plugin
+// before starting the server. This allows the plugin to make RPC calls back to the Icinga Notifications process.
+type RPCEndpointReceiver interface {
+	ReceiveEndpoint(ctx context.Context, ep *jsonrpc.Endpoint)
```
oxzi, line 209: Unless I missed something, this interface is neither tested in this PR nor used in https://github.com/Icinga/icinga-notifications/pull/473 (at the time of writing). So, please take the following with a grain of salt, as it is all based on assumptions.

Following the prior internal mutex fun, giving a plugin access to the internal `Endpoint` enables the plugin to mess with this as well. At the moment, there are three methods in the `Plugin` interface, where I would expect that - unless for some background job - a JSONRPC call is most likely made from `Plugin.SendNotification`.

At the moment, `SendNotification` is called from `rpcHandler.Handle`, which is passed as the `jsonrpc2.Handler` via `plugin.Run` into `jsonrpc.New`.

https://github.com/Icinga/icinga-go-library/blob/07122f2f2fef0ab442f2d055dc0545c32fdc919d/notifications/plugin/plugin.go#L279

For `jsonrpc2.Conn`, the handler is called synchronously (!) in `Conn.readMessage`

https://github.com/sourcegraph/jsonrpc2/blob/v0.2.1/conn.go#L205

If a `Plugin` implementation now does an `Endpoint.Call` and waits for the reply within `Plugin.SendNotification`, this will deadlock as `Conn.readMessage` sits at the handler call and cannot process more messages.

However, I guess the easiest fix would be to use `jsonrpc.AsyncHandler`. This is even suggested in the docs:

	// Handle is called to handle a request. No other requests are handled
	// until it returns. If you do not require strict ordering behavior
	// of received RPCs, it is suggested to wrap your handler in
	// AsyncHandler.

<https://pkg.go.dev/github.com/sourcegraph/jsonrpc2#Handler>

How about just wrapping `h` in `jsonrpc.New` and document in `plugin.Plugin` that the methods might get called concurrently. I guess this was even the case before this change with the old code.

notifications/jsonrpc/jsonrpc.go
yhabteab: Changed the parameter to be a zap sugared logger instead. If one is provided, then no errors will be forwared to peer.

notifications/plugin/plugin.go
yhabteab: > How about just wrapping `h` in `jsonrpc.New` and document in `plugin.Plugin` that the methods might get called concurrently. I guess this was even the case before this change with the old code.

Yes, I already considered in using the async handler but since Icinga Notifications won't use the plugin concurrently anymore (due to the HA related changes made recently - processing only a single event at a time), I opt to not using it. But given that we don't have control over all the plugins and what they might be doing using the async handler is safer. So changed it.

notifications/jsonrpc/jsonrpc.go
yhabteab: > and make `logsCh` buffered - dunno, 64 entries - and add a `default` case to bail out even if the buffered channel is full.

Done. And I'm now using ctx with timeout of 2s instead.

yhabteab: > So, at least the API doc regarding the context cancellation is wrong.

`loggerFn` isn't the only user of the `LogNotify` method, so no it was only wrong in this specific case since there was no context cancellation due to the ctx background.

go.sum
```diff
@@ -58,6 +60,10 @@ github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZb
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
 github.com/redis/go-redis/v9 v9.21.0 h1:FPBE4hhbAke+TLmcY3WkpbDffJEomdqPn3HYiqAtL9E=
 github.com/redis/go-redis/v9 v9.21.0/go.mod h1:v/M13XI1PVCDcm01VtPFOADfZtHf8YW3baQf57KlIkA=
+github.com/sourcegraph/jsonrpc2 v0.2.1 h1:2GtljixMQYUYCmIg7W9aF2dFmniq/mOr2T9tFRh6zSQ=
+github.com/sourcegraph/jsonrpc2 v0.2.1/go.mod h1:ZafdZgk/axhT1cvZAPOhw+95nz2I/Ra5qMlU4gTRwIo=
```
oxzi: `go mod tidy`

Seems like this check is missing in the IGL's CI.

notifications/jsonrpc/jsonrpc.go
```diff
@@ -0,0 +1,208 @@
...
+		Message string        `json:"message"`
+		Fields  []any         `json:"fields"`
+	}
+)
+
+// New creates a new JSON-RPC endpoint with the given context, read and write streams, and request handler.
+//
+// If the handler is nil, the endpoint will function as a client-only endpoint that can send requests or push
+// notifications but will not handle incoming RPC requests. The read and write streams are used for communication
+// with the other side of the JSON-RPC connection, and might be connected to a pipe, or any other I/O stream.
```
oxzi: Sorry, I have not considered this detail before. At the moment, if `h` is `nil`, a new `jsonrpc2.AsyncHandler` will be created, wrapping a `nil` pointer. If this endpoint will receive a message regardless of its "send only" character, it will panic.

- https://github.com/sourcegraph/jsonrpc2/blob/v0.2.2/async.go#L8
- https://github.com/sourcegraph/jsonrpc2/blob/v0.2.2/async.go#L16

go.sum
yhabteab: Fixed.

notifications/jsonrpc/jsonrpc.go
yhabteab: For its panicing behavior, there's nothing I can do because that's how the library works. If you send RPC calls to peer that doesn't have a registered handler, then it will panic internally, but I've added an explicit comment describing this.

PR #227, opened 2026-08-03
config.Load: Fix Validation for Password Files

oxzi:
No longer return an error for config.LoadPasswordFile if both the password and the password file are set. This error was only nannyism.

The config.LoadPasswordFile function sets the password file and errors when both the password and password file are set. All this is documented behavior. The function is used as a helper in Validator interface implementation, e.g., in database.Config.

When invoked via config.Load, the Validate method is called twice. First, for the YAML configuration via config.FromYAMLFile, then, for the environment variables via config.FromEnv.

When the YAML configuration contains a password file, the config.FromYAMLFile call will populate the password field. Then, the subsequent config.FromEnv call fails, as the second config.LoadPasswordFile will find both a password file as well as a password, being set in the previous call.

This error existed since the function was introduced, but, as it seems, it was only ever tested with environment variables. Note, this error does not unveil when using environment variables, as there the first call is a no op.

---

In Icinga DB, try the following `database` config - only having `password_file` set, no `password`:

```
database:
  type: mysql
  host: localhost
  database: icingadb
  user: icingadb
  password_file: tmp-pass
```

And start it with a `tmp-pass` file:

```
$ echo a > tmp-pass
$ ./icingadb --config config.yml
invalid configuration: invalid database configuration: both password and password file are set
```

---

This is how the tests would fail without the change to `LoadPasswordFile`.

```
--- FAIL: TestLoad (0.00s)
    --- FAIL: TestLoad/Password_file_in_YAML (0.00s)
        testutils.go:41:
                Error Trace:    /home/apenning/git/github.com/Icinga/icinga-go-library/testutils/testutils.go:41
                Error:          Received unexpected error:
                                invalid configuration: both password and password file are set
                                github.com/icinga/icinga-go-library/config.Load
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/config/config.go:298
                                github.com/icinga/icinga-go-library/config.TestLoad.func1.1
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/config/config_test.go:453
                                github.com/icinga/icinga-go-library/testutils.WithYAMLFile
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/testutils/testutils.go:97
                                github.com/icinga/icinga-go-library/config.TestLoad.func1
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/config/config_test.go:452
                                github.com/icinga/icinga-go-library/config.TestLoad.TestCase[...].F.func7
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/testutils/testutils.go:36
                                testing.tRunner
                                        /nix/store/rcz9i4msbg3178grqll9h98b07dwk7zg-go-1.26.5/share/go/src/testing/testing.go:2036
                                runtime.goexit
                                        /nix/store/rcz9i4msbg3178grqll9h98b07dwk7zg-go-1.26.5/share/go/src/runtime/asm_amd64.s:1771
                Test:           TestLoad/Password_file_in_YAML
    --- FAIL: TestLoad/Password_file_in_both (0.00s)
        testutils.go:41:
                Error Trace:    /home/apenning/git/github.com/Icinga/icinga-go-library/testutils/testutils.go:41
                Error:          Received unexpected error:
                                invalid configuration: both password and password file are set
                                github.com/icinga/icinga-go-library/config.Load
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/config/config.go:298
                                github.com/icinga/icinga-go-library/config.TestLoad.func1.1
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/config/config_test.go:453
                                github.com/icinga/icinga-go-library/testutils.WithYAMLFile
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/testutils/testutils.go:97
                                github.com/icinga/icinga-go-library/config.TestLoad.func1
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/config/config_test.go:452
                                github.com/icinga/icinga-go-library/config.TestLoad.TestCase[...].F.func7
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/testutils/testutils.go:36
                                testing.tRunner
                                        /nix/store/rcz9i4msbg3178grqll9h98b07dwk7zg-go-1.26.5/share/go/src/testing/testing.go:2036
                                runtime.goexit
                                        /nix/store/rcz9i4msbg3178grqll9h98b07dwk7zg-go-1.26.5/share/go/src/runtime/asm_amd64.s:1771
                Test:           TestLoad/Password_file_in_both
--- FAIL: TestLoadPasswordFile (0.00s)
    --- FAIL: TestLoadPasswordFile/password_file_only (0.00s)
        config_test.go:624:
                Error Trace:    /home/apenning/git/github.com/Icinga/icinga-go-library/config/config_test.go:624
                Error:          Received unexpected error:
                                both password and password file are set
                                github.com/icinga/icinga-go-library/config.LoadPasswordFile
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/config/config.go:368
                                github.com/icinga/icinga-go-library/config.TestLoadPasswordFile.func1
                                        /home/apenning/git/github.com/Icinga/icinga-go-library/config/config_test.go:624
                                testing.tRunner
                                        /nix/store/rcz9i4msbg3178grqll9h98b07dwk7zg-go-1.26.5/share/go/src/testing/testing.go:2036
                                runtime.goexit
                                        /nix/store/rcz9i4msbg3178grqll9h98b07dwk7zg-go-1.26.5/share/go/src/runtime/asm_amd64.s:1771
                Test:           TestLoadPasswordFile/password_file_only
                Messages:       second call must not fail
FAIL
FAIL    github.com/icinga/icinga-go-library/config      3.144s
FAIL
```


config/config.go
```diff
@@ -361,13 +361,8 @@ func validateNonNilStructPointer(v any) error {
...
 //
 // This is a helper function to be used in Validator.Validate implementations on types where
-// both a password and a password file field are available. If both a password and a password
-// file are given, an error is returned.
+// both a password and a password file field are available.
 func LoadPasswordFile(password *string, passwordFile string) error {
-	if *password != "" && passwordFile != "" {
-		return errors.New("both password and password file are set")
-	}
-
```
yhabteab, line 370: This now breaks [users that relied on it](https://github.com/Icinga/icinga-notifications/blob/main/config.example.yml#L31-L33). Now, there is no way to tell which option has higher priority as setting both options causes `password_file` to always win regardless if it's from ENV or YAML file. The actual problem is that the `Validate` method is called twice unnecessarily. Either we should just skip environment parsing when `FromYAMLFile` passes the validation or `FromYAMLFile` and `FromEnv` should both skip the `Validate` call when invoked from `config.Load`, and `config.Load` makes a final `v.Validate()` call after calling both the other functions instead.

oxzi: This was not part of any release so far: https://github.com/Icinga/icinga-notifications/blob/v0.2.0/config.example.yml

We can either update the docs to state that the `password_file` takes precedence or I can rework this PR to use a `sync.Once`. What would you prefer? (I am fine with both.)

yhabteab: > We can either update the docs to state that the `password_file` takes precedence or I can rework this PR to use a `sync.Once`. What would you prefer?

Just for completeness, so it doesn't appear this is waiting for a reply from me. As discussed offline, I've actually suggested what we could do instead in my initial comment. So, I would just do that and call `Validate` only once, which should be trivial to implement.

oxzi: Updated, feel free to take another look.

